PR #1, opened 2023-10-04
Fix addon imports and name

antirotor:
This is fixing underscore in module name to make it importable and replacing imports of kind `from openpype.hosts.resolve import api` -> `from ayon_resolve import api`.


client/ayon_resolve/plugins/publish/extract_workfile.py
```diff
@@ -2,7 +2,8 @@
 import pyblish.api
 
 from openpype.pipeline import publish
-from openpype.hosts.resolve.api.lib import get_project_manager
+from ayon_resolve.api.lib import get_project_manager
```
iLLiCiTiT, line 5:
```suggestion
```

PR #9, opened 2024-07-08
Do not create timeline on opening Loader UI

BigRoy:
### Changes

Do not create timeline when opening the Loader UI or on calling `get_current_timeline_items`

### Testing notes

1. Loading clips and media should still work, even when no timeline currently exists in project.
2. Manager should still work, even when no timeline currently exists in project.
3. Loading clip should still create timeline when loading into timeline.


jakubjezek001 (approved):
perfect solution, thanks!

PR #11, opened 2024-07-18
Resolve: enhancing editorial pkg loader

jakubjezek001:
We've upgraded the loader in several ways:

*   The timeline now includes a `VFX data` JSON string with Load and Publish metadata.
*   Each timeline is randomly assigned one of the available clip colors.
*   The update function ensures that the newest version is loaded into the media pool.
*   Timelines are organized into hierarchical bins, each with a dedicated version bin.
*   Manager Tool can easily list timelines within containers.

# Testing steps
1. Use Traypublisher and publish couple times Editorial Package folder here is testing of ynput internal files (sensitive data) https://drive.google.com/open?id=1XbYdnblAuYuCYPaEBd48iSHhbcfjiOPm&usp=drive_fs
2. Open Resolve and Loader
3. Load the editorial_pkg product in lower version so you can version manage via Manager
4. After it is loaded and version updated notice there are multiple version folders and each of timeline is having different clip color assigned.

### DIFF
--- a/client/ayon_resolve/api/lib.py
+++ b/client/ayon_resolve/api/lib.py
@@ -1,7 +1,8 @@
 import sys
-import json
 import re
 import os
+import json
+import uuid
 import contextlib
 from typing import List, Dict, Any
 from opentimelineio import opentime
@@ -38,6 +39,27 @@ self.temp_marker_frame = None
 
 # OpenPype default timeline
 self.pype_timeline_name = "OpenPypeTimeline"
+
+
+def get_timeline_media_pool_item(timeline, root=None):
+    """Return MediaPoolItem from Timeline"""
+
+    # Due to limitations in the Resolve API we can't get
+    # the media pool item directly from the timeline.
+    # We can find it by name, however names are not
+    # enforced to be unique across bins. So, we give it
+    # unique name.
+    original_name = timeline.GetName()
+    identifier = str(uuid.uuid4().hex)
+    try:
+        timeline.SetName(identifier)
+        for item in iter_all_media_pool_clips(root=root):
+            if item.GetName() != identifier:
+                continue
+            return item
+    finally:
+        # Revert to original name
+        timeline.SetName(original_name)
 
 
 @contextlib.contextmanager
--- a/client/ayon_resolve/api/lib.py
+++ b/client/ayon_resolve/api/lib.py
@@ -961,7 +961,8 @@ def iter_all_media_pool_clips(root=None):
     """Recursively iterate all media pool clips in current project
 
     Args:
-        root (resolve.Folder)[optional]: root folder / bin object
+        root (Optional[resolve.Folder]): root folder / bin object.
+            When None, defaults to media pool root folder.
     """
     root = root or get_current_project().GetMediaPool().GetRootFolder()
     queue = [root]
--- a/client/ayon_resolve/api/lib.py
+++ b/client/ayon_resolve/api/lib.py
@@ -960,9 +960,13 @@ def get_reformated_path(path, padded=False, first=False):
     return path
 
 
-def iter_all_media_pool_clips():
-    """Recursively iterate all media pool clips in current project"""
-    root = get_current_project().GetMediaPool().GetRootFolder()
+def iter_all_media_pool_clips(root=None):
+    """Recursively iterate all media pool clips in current project
+
+    Args:
+        root (resolve.Folder)[optional]: root folder / bin object
+    """
+    root = root or get_current_project().GetMediaPool().GetRootFolder()
     queue = [root]
     for folder in queue:
         for clip in folder.GetClipList():
--- a/client/ayon_resolve/api/lib.py
+++ b/client/ayon_resolve/api/lib.py
@@ -41,8 +41,17 @@ self.temp_marker_frame = None
 self.pype_timeline_name = "OpenPypeTimeline"
 
 
-def get_timeline_media_pool_item(timeline, root=None):
-    """Return MediaPoolItem from Timeline"""
+def get_timeline_media_pool_item(timeline, root=None) -> object:
+    """Return MediaPoolItem from Timeline
+
+
+    Args:
+        timeline (resolve.Timeline): timeline object
+        root (resolve.Folder): root folder / bin object
+
+    Returns:
+        resolve.MediaPoolItem: media pool item from timeline
+    """
 
     # Due to limitations in the Resolve API we can't get
     # the media pool item directly from the timeline.
--- a/client/ayon_resolve/api/pipeline.py
+++ b/client/ayon_resolve/api/pipeline.py
@@ -157,6 +157,9 @@ def ls():
         data = json.loads(data)
 
         # treat data as container
+        # There might be cases where clip's metadata are having additional
+        # because it needs to store 'load' and 'publish' data. In that case
+        # we need to get only 'load' data
         if data.get("load"):
             data = data["load"]
 
--- a/client/ayon_resolve/api/pipeline.py
+++ b/client/ayon_resolve/api/pipeline.py
@@ -156,6 +156,10 @@ def ls():
             continue
         data = json.loads(data)
 
+        # treat data as container
+        if data.get("load"):
+            data = data["load"]
+
         # If not all required data, skip it
         required = ['schema', 'id', 'loader', 'representation']
         if not all(key in data for key in required):
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -5,6 +5,7 @@ import copy
 import qargparse
 from qtpy import QtWidgets, QtCore
 
+from ayon_core.pipeline.constants import AVALON_INSTANCE_ID
 from ayon_core.settings import get_current_project_settings
 from ayon_core.pipeline import (
     LegacyCreator,
@@ -899,6 +900,35 @@ class PublishClip:
         for key in par_split:
             parent = self._convert_to_entity(key)
             self.parents.append(parent)
+
+
+def get_editorial_publish_data(
+    folder_path,
+    product_name,
+    version=None
+) -> dict:
+    """Get editorial publish data from context.
+
+    Args:
+        folder_path (str): Folder path where editorial package is located.
+        product_name (str): Editorial product name.
+        version (Optional[str]): Editorial product version. Defaults to None.
+
+    Returns:
+        dict: Editorial publish data.
+    """
+    data = {
+        "id": AVALON_INSTANCE_ID,
+        "productType": "editorial_pkg",
+        "productName": product_name,
+        "folderPath": folder_path,
+        "active": True,
+    }
+
+    if version:
+        data["version"] = version
+
+    return data
 
 
 def get_representation_files(representation):
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -16,22 +16,13 @@ class CreateEditorialPackage(LegacyCreator):
     def process(self):
         """Process the creation of the editorial package."""
         current_timeline = lib.get_current_timeline()
-        timeline_name = current_timeline.GetName()
 
-        # get timeline media pool item for metadata update
-        timeline_media_pool_item = None
-        for item in lib.iter_all_media_pool_clips():
-            item_name = item.GetName()
-            if item_name != timeline_name:
-                continue
-            timeline_media_pool_item = item
-            break
+        timeline_media_pool_item = lib.get_timeline_media_pool_item(
+            current_timeline
+        )
 
-        # Update the metadata
-        if timeline_media_pool_item:
-            publish_data = {"publish": self.data}
-            timeline_media_pool_item.SetMetadata(
-                lib.pype_tag_name, json.dumps(publish_data)
-            )
-        else:
-            NameError("Timeline not found in media pool.")
+        publish_data = {"publish": self.data}
+
+        timeline_media_pool_item.SetMetadata(
+            lib.pype_tag_name, json.dumps(publish_data)
+        )
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -1,4 +1,3 @@
-from pprint import pformat
 import json
 from ayon_core.pipeline.create.legacy_create import LegacyCreator
 
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -33,3 +33,5 @@ class CreateEditorialPackage(LegacyCreator):
             timeline_media_pool_item.SetMetadata(
                 lib.pype_tag_name, json.dumps(publish_data)
             )
+        else:
+            NameError("Timeline not found in media pool.")
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -15,16 +15,6 @@ class CreateEditorialPackage(LegacyCreator):
 
     def process(self):
         """Process the creation of the editorial package."""
-        if (self.options or {}).get("useSelection"):
-            # if selected is on then get only timeline from media pool
-            # selection
-            self.convert_timeline_to_editorial_package_instance()
-        else:
-            # if selected is on then get only active timeline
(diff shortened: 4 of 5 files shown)


client/ayon_resolve/api/lib.py
```diff
@@ -960,9 +960,13 @@ def get_reformated_path(path, padded=False, first=False):
...
 
 
-def iter_all_media_pool_clips():
-    """Recursively iterate all media pool clips in current project"""
-    root = get_current_project().GetMediaPool().GetRootFolder()
+def iter_all_media_pool_clips(root=None):
+    """Recursively iterate all media pool clips in current project
+
+    Args:
+        root (resolve.Folder)[optional]: root folder / bin object
```
BigRoy, line 964:
```suggestion
        root (Optional[resolve.Folder]): root folder / bin object.
            When None, defaults to media pool root folder.
```

client/ayon_resolve/api/pipeline.py
```diff
@@ -156,6 +156,10 @@ def ls():
             continue
         data = json.loads(data)
 
+        # treat data as container
+        if data.get("load"):
+            data = data["load"]
```
BigRoy, line 161: Not sure about this - why do we need this. I feel this comment is also lacking as to for what use case this is. Might be good to elaborate.

client/ayon_resolve/api/plugin.py
```diff
@@ -901,6 +902,35 @@ def _create_parents(self):
...
+        "productType": "editorial_pkg",
+        "productName": product_name,
+        "folderPath": folder_path,
+        "active": True,
+    }
+
+    if version:
+        data["version"] = version
+
+    return data
```
BigRoy, line 933: This feels a bit weird in `plugin.py`. Should this be in `lib.py` or in a dedicated module, e.g. `ayon_resolve.api.editorial` or whatever?

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -0,0 +1,46 @@
...
+
+    def process(self):
+        """Process the creation of the editorial package."""
+        if (self.options or {}).get("useSelection"):
+            # if selected is on then get only timeline from media pool
+            # selection
+            self.convert_timeline_to_editorial_package_instance()
+        else:
+            # if selected is on then get only active timeline
+            self.convert_timeline_to_editorial_package_instance()
```
BigRoy, line 24: Am I blind or is the if and else statement the same? Something is wrong here?

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -0,0 +1,46 @@
...
+                continue
+            timeline_media_pool_item = item
+            break
+
+        # Update the metadata
+        if timeline_media_pool_item:
+            publish_data = {"publish": self.data}
+            timeline_media_pool_item.SetMetadata(
+                lib.pype_tag_name, json.dumps(publish_data)
+            )
```
BigRoy, line 45: Should this log a warning if it failed to do anything? (e.g. the else case?)

Or actually... shouldn't it ALWAYS have a media pool item? Like, remove the if branch or raise an error if it's not found for the current timeline?

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -0,0 +1,46 @@
...
+        timeline_name = current_timeline.GetName()
+
+        # get timeline media pool item for metadata update
+        timeline_media_pool_item = None
+        for item in lib.iter_all_media_pool_clips():
+            item_name = item.GetName()
+            if item_name != timeline_name:
+                continue
+            timeline_media_pool_item = item
+            break
```
BigRoy, line 38: Uh... are you trying to do this?
```suggestion
        timeline_media_pool_item = current_timeline.GetMediaPoolItem()
```

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -0,0 +1,46 @@
...
+            # selection
+            self.convert_timeline_to_editorial_package_instance()
+        else:
+            # if selected is on then get only active timeline
+            self.convert_timeline_to_editorial_package_instance()
+
+    def convert_timeline_to_editorial_package_instance(self):
+        """Convert timeline to editorial package instance."""
+        current_timeline = lib.get_current_timeline()
+        timeline_name = current_timeline.GetName()
```
BigRoy, line 29: This will error if there's no current timeline.
```suggestion
        current_timeline = lib.get_current_timeline()
        if not current_timeline:
            raise RuntimeError("Make sure to have an active current timeline.")
        timeline_name = current_timeline.GetName()
```

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,126 @@ def load(self, context, name, namespace, data):
...
         timeline = media_pool.ImportTimelineFromFile(files, import_options)
+
+        # get timeline media pool item for metadata update
+        timeline_media_pool_item = None
+        for item in lib.iter_all_media_pool_clips(root=loaded_bin):
+            item_name = item.GetName()
+            if item_name != loaded_timeline_name:
+                continue
+            timeline_media_pool_item = item
+            break
```
BigRoy, line 61:
```suggestion
        timeline_media_pool_item = timeline.GetMediaPoolItem()
```

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,126 @@ def load(self, context, name, namespace, data):
...
+
+        # Update the metadata
+        if timeline_media_pool_item:
```
BigRoy, line 64: Shouldn't the timeline media pool item always exist? What am I missing? This should either error if not found... or?

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,126 @@ def load(self, context, name, namespace, data):
...
+        data["publish"] = get_editorial_publish_data(
+            folder_path=context["folder"]["path"],
+            product_name=context["product"]["name"],
+            version=version_entity["version"],
+        )
+
+        return data
+
+    def get_clip_color(self):
+        """Return clip color based on version data."""
```
BigRoy, line 138: Docstring is wrong. It doesn't base it on anything. Currently this is best of renamed to `get_random_clip_color()`

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,126 @@ def load(self, context, name, namespace, data):
...
+            clip_data = self._get_container_data(
+                context, data)
+
+            timeline_media_pool_item.SetMetadata(
+                lib.pype_tag_name, json.dumps(clip_data)
+            )
+
+            # set clip color based on random choice
+            clip_color = self.get_clip_color()
+            print("Clip color: ", clip_color)
```
BigRoy, line 74: Should this be `self.log.info` calls?

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,126 @@ def load(self, context, name, namespace, data):
...
+            timeline_media_pool_item.SetClipColor(clip_color)
+
+        # TODO: there are two ways to import timeline resources (representation
+        #   and resources folder) but Resolve seems to ignore any of this
+        #   since it is importing sources automatically. But we might need
+        #   to at least set some metadata to those loaded media pool items
         print("Timeline imported: ", timeline)
```
BigRoy, line 81: Should this be `self.log.info` calls?

BigRoy (changes requested):

BigRoy:
Could you provide testing instruction (or link me to a docs page) on how to generate an Editorial Pkg? And what to test for in this PR when loading?

client/ayon_resolve/api/plugin.py
jakubjezek001: This code is related to plugin related 'tool-set' so it seems to me it is ok to be added here. There is also `get_representation_files` in the plugin. This function is used by creator and loader plugin only.

client/ayon_resolve/api/pipeline.py
jakubjezek001: I hope it makes more sense now 64071b6cafaf7ec724ffea01890937ed6b8fa206

client/ayon_resolve/plugins/load/load_editorial_package.py
jakubjezek001: yeah I wish so, but it does not log anything so print is more safe.

client/ayon_resolve/plugins/load/load_editorial_package.py
jakubjezek001: **this** will result in:
```
  File "C:\CODE\__PYPE\ayon-core\client\ayon_core\tools\loader\models\actions.py", line 740, in _load_representations_by_loader
    load_with_repre_context(
  File "C:\CODE\__PYPE\ayon-core\client\ayon_core\pipeline\load\utils.py", line 325, in load_with_repre_context
    return loader.load(repre_context, name, namespace, options)
  File "C:\CODE\__PYPE\ayon-resolve\client\ayon_resolve\plugins\load\load_editorial_package.py", line 55, in load
    timeline_media_pool_item = timeline.GetMediaPoolItem()
TypeError: 'NoneType' object is not callable
```

Sadly *Timeline* object is having different set of attributes then *TimelineItem*.

client/ayon_resolve/plugins/create/create_editorial_package.py
jakubjezek001: done

client/ayon_resolve/plugins/load/load_editorial_package.py
jakubjezek001: improved

client/ayon_resolve/plugins/load/load_editorial_package.py
jakubjezek001: done

client/ayon_resolve/plugins/load/load_editorial_package.py
BigRoy: Doesn't that just mean that maybe we need to configure a custom logging handler for Resolve that maybe ends up just doing the print calls?

client/ayon_resolve/plugins/load/load_editorial_package.py
BigRoy: Ah of course! 

The tricky thing is that the name isn't entirely enforced to be unique unfortunately:
- https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=177766
- https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=178493

Both list the same issue. (Note that the name not being unique may also be due to a 'bug' but apparently there are ways to get resolve into that state)

Anyway - we might want to at least make this a dedicated lib function and ensure what we find is what we intend to find? How about this in `ayon_resolve.api.lib`:

```python
import uuid

def get_timeline_media_pool_item(timeline, root=None):
    """Return MediaPoolItem from Timeline"""
    
    name = timeline.GetName()
    
    # Due to limitations in the Resolve API we can't get
    # the media pool item directly from the timeline.
    # We can find it by name, however - but names are not
    # enforced to be unique across bins. So, we give it
    # unique name.
    identifier = str(uuid.uuid4().hex)
    try:
        timeline.SetName(identifier)
        for item in iter_all_media_pool_clips(root=root):
            item.GetName() == identifier:
                return item
        
    finally:
        # Revert to original name
        timeline.SetName(name)
        if item_name != loaded_timeline_name:
            continue
```

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,125 @@ def load(self, context, name, namespace, data):
...
+
+        timeline_media_pool_item.SetMetadata(
+            lib.pype_tag_name, json.dumps(clip_data)
+        )
+
+        # set clip color based on random choice
+        clip_color = self.get_clip_color()
+        timeline_media_pool_item.SetClipColor(clip_color)
+
+
```
BigRoy, line 75:
```suggestion
```

client/ayon_resolve/plugins/load/load_editorial_package.py
jakubjezek001: thanks a lot for the suggestion, it is implemented

jakubjezek001:
@BigRoy have you had a chance to test it?

client/ayon_resolve/plugins/create/create_editorial_package.py
jakubjezek001: reimplemented the function from lib `get_timeline_media_pool_item`

LiborBatek (changes requested):
I have followed the testing steps and  I have succesfully created that editorial package from the provided data (link) and tried to ```Load``` inside Resolve.

It loaded correctly but for some reason all clips are offline and cannot be viewed:

Screenshot 2024-08-05 160208

even thought when inspecting published version of the editorial package each version having resource file with movies...


Screenshot 2024-08-05 160221

There is also another issue when performing ```Manage > Set Version``` that the already loaded editorial package stays listed in the ```Scene Inventory``` resulting into 2 entries e.g. ```v001``` and also ```v003``` - which should happen.

jakubjezek001:
> I have followed the testing steps and I have succesfully created that editorial package from the provided data (link) and tried to `Load` inside Resolve.

This seems to be doing what it should - judging from the screenshot, but the issue with offline clips is the version of Resolve. I will try to include the case.

LiborBatek (approved):

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -32,21 +36,127 @@ def load(self, context, name, namespace, data):
...
+        """Return metadata related to the representation and version."""
+
+        # add additional metadata from the version to imprint AYON knob
+        version_entity = context["version"]
+
+        # remove unnecessary keys from the data
+        for key in ["_item", "name"]:
+            if key not in data:
+                continue
+            data.pop(key)
```
robin-ynput, line 106:
```suggestion
        for key in ("_item", "name"):
            data.pop(key, None)  # remove unnecessary key from the data if it exists
```

jakubjezek001:
> Just to flag, if I understand correctly these PR #11 et #12 will introduce changes based on the legacy creator/publisher API. How do you forsee the merge ? This stuff firstly then the new creator or the other way around ?
> 
> (I totally don't mind doing the diff btw)

Apologies if this is a bit confusing. This needs to be merged first, and then the target of PR #12 will automatically change to develop. The dependency is mentioned in the description of #12. I hope I understood your question correctly.

client/ayon_resolve/api/lib.py
```diff
@@ -40,6 +41,36 @@
...
+        resolve.MediaPoolItem: media pool item from timeline
+    """
+
+    # Due to limitations in the Resolve API we can't get
+    # the media pool item directly from the timeline.
+    # We can find it by name, however names are not
+    # enforced to be unique across bins. So, we give it
+    # unique name.
+    original_name = timeline.GetName()
+    identifier = str(uuid.uuid4().hex)
```
iLLiCiTiT, line 62:
```suggestion
    identifier = uuid.uuid4().hex
```

client/ayon_resolve/api/lib.py
```diff
@@ -40,6 +41,36 @@
...
+    try:
+        timeline.SetName(identifier)
+        for item in iter_all_media_pool_clips(root=root):
+            if item.GetName() != identifier:
+                continue
+            return item
```
iLLiCiTiT, line 68:
```suggestion
            if item.GetName() == identifier:
                return item
```

client/ayon_resolve/plugins/load/load_editorial_package.py
```diff
@@ -1,11 +1,15 @@
+import json
 from pathlib import Path
+import random
 
 from ayon_core.pipeline import (
+    AVALON_CONTAINER_ID,
```
iLLiCiTiT, line 6: You should use `AYON_CONTAINER_ID`, but looking in the code, you coudl not add it to the metadata at all, and just make sure `"id"` is filled when you return in `list_containers`. The same for `AYON_INSTANCE_ID` btw.

iLLiCiTiT: And also use `ayon:container-3.0` instead of `openpype:container-2.0`.

PR #12, opened 2024-07-18
Resolve: publishing editorial pkg from active timeline

### LINKED ISSUE #19: AY-5638_resolve: publishing timeline
### Please describe the feature you have in mind and explain what the current shortcomings are?

Choose a timeline (it must have metadata from when we loaded it), Clients don't need to take any extra steps, such as setting up a timeline instance.
Open the publisher, and the instance of the timeline should appear.
Publishing process must turn the timeline into one video - we have to set up a rendering job in Resolve and wait for the file. Then we can publish along with an OTIO file made from the timeline.
Burn-ins should probably be added dynamically to the Resolve timeline. Currently, we don't support multi-contextual burn-ins through our own FFmpeg-based plugin.
Burn-ins should be added to each clip, displaying the name, clip time-code, timeline time-code, timeline name, and editor's name.

### How would you imagine the implementation of the feature?
_No response_

### Describe alternatives you've considered:
_No response_

### Additional context:

_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-5638](https://app.clickup.com/t/86bz01dma)_

### LINKED ISSUE #25: AY-6300_Resolve: publishing full timeline
### Please describe the enhancement you have in mind and explain what the current shortcomings are?

Hi guys, in Davinci resolve, could we have a Render instance that could render out the whole timeline video with audio? (edited)

### How would you imagine the implementation of the enhancemenent?
_No response_

### Describe alternatives you've considered:
_No response_

### Additional context:
()
_(might be a private channel)_

_This issue was automatically created from Clickup ticket [AY-6300](https://app.clickup.com/t/86bzw3cmd)_

jakubjezek001:
We've upgraded the loader in several ways:

*   Collect all instances of timelines labeled as "editorial package" from the mediapool. Those could be loaded timelines via dedicated loader for editorial package or created via Creator.

This PR is based on #11

closes #25 
closes #19 

## Testing steps
- select any timeline with clips and use Create with `Editorial Package` product type
- context selector window should popup 
- select folder and task context which should be used for the product
- close Creator and open Publisher
- publish and all should look as expected - editorial pkg product is now ready to be loaded.


client/ayon_resolve/plugins/publish/extract_editorial_package.py
```diff
@@ -0,0 +1,190 @@
...
+                    clip.media_reference = new_media_reference
+
+                    # replace clip source range
+                    new_source_range = otio.opentime.TimeRange(
+                        start_time=otio.opentime.RationalTime(
+                            value=current_end_frame, rate=timeline_fps
+                        ),
+                        duration=clip.source_range.duration,
+                    )
+                    clip.source_range = new_source_range
```
robin-ynput, line 127: I'm not sure I understand why the clip source range needs to be edited ?
Is it because swaping the media reference reset the source range (and if so isn't `new_source_range` always equal initial `clip.range_in_parent()` ?

client/ayon_resolve/api/lib.py
```diff
@@ -94,17 +94,60 @@ def maintain_current_timeline(to_timeline: object,
...
+        >>> with open_specific_page_by_name("Deliver"):
+        ...     print("Deliver page is open")
+        Deliver page is open
+    """
+    from . import bmdvr
+    current_page = bmdvr.GetCurrentPage()
+    log.info(f"Current page: {current_page}")
+
+    log.info(f"Opening page: {page_name}")
+    bmdvr.OpenPage(page_name)
```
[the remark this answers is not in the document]
BigRoy: I believe the Resolve docs actually mention that the return value from the function call usually represent success or failure? Anyway, good catch.

I do feel like the function name itself doesn't really hint at all that it's a context manager? (Should it?)
Aside of that, the logging itself feels bit overkill/redundant. But that was maybe for debugging that "page did not change" issue that @robin-ynput pointed out?

client/ayon_resolve/api/lib.py
```diff
@@ -1001,3 +1044,38 @@ def iter_all_media_pool_clips(root=None):
...
+            # use native otio export
+            export_timeline_otio(timeline, filepath)
+
+    Args:
+        timeline (resolve.Timeline): resolve's object
+        filepath (str): otio file path
+
+    Returns:
+        bool: True if success
+    """
```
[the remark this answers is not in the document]
jakubjezek001: good point, I am renaming it to `export_timeline_otio_native`

client/ayon_resolve/api/rendering.py
```diff
@@ -0,0 +1,240 @@
...
+        log.error(f"Failed to render timelines: {timeline_name}")
+        return False
+
+    wait_for_rendering_completion()
+
+    delete_all_processed_jobs()
+
+    log.info("Rendering is completed.")
+
+    return True
```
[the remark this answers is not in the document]
jakubjezek001: I like it, thanks for the suggestion. It had been committed here 64a6d290755a17e7c56342e980112172d1182d2e

client/ayon_resolve/plugins/publish/extract_editorial_package.py
jakubjezek001: The issue is that we are relinking the original individual media sources and their source ranges onto one media source rendered from the timeline. There might be an easier solution you can suggest. This was the best I could find at the moment.

robin-ynput: I understand better yes. 
If the swapped media is the one rendered from the timeline, then media clip source range should always `clip.range_in_parent()`. This might save you some computation here.

client/ayon_resolve/plugins/publish/extract_editorial_package.py
```diff
@@ -0,0 +1,190 @@
...
+
+    TODO:
+        - find reference video representation
+        - figure out target publishing rootless path for reference paths
+        - swap all references file names in OTIO file point at the rootless path
+    """
+
+    label = "Extract Editorial Package"
+    order = pyblish.api.ExtractorOrder + 0.45
+    families = ["editorial_pkg"]
```
robin-ynput, line 29: I tried to publish/load an `editorial_pkg` but unfortunately it did not work for me straight on.
resolve_failed_integration

Investigating a bit, I realized neither `Extract Editorial Package` nor `Extract Intermediate Representation` were part of my plugins loop.
resolve_missing_plugins

Investigating a bit, I realized a `families` mismatch. The instance data I got out of the creator is:
```
Editorial Package: {'family': 'default', 'name': 'TimelineImg', 'id': 'pyblish.avalon.instance', 'productType': 'editorial_pkg', 'folderPath': '/Root', 'productName': 'editorial_pkgMain', 'active': True, 'variant': 'Main', 'taskId': '415050805f1111ef93b231cb88eabe44', 'taskName': 'Generic', 'mediaPoolItem': <BlackmagicFusion.PyRemoteObject object at 0x00000000054DAE70>, 'taskEntity': {'attrib': {'resolutionHeight': 1080, 'frameEnd': 1001, 'resolutionWidth': 1920, 'handleStart': 0, 'endDate': None, 'clipOut': 1, 'frameStart': 1001, 'tools': None, 'pixelAspect': 1.0, 'clipIn': 1, 'handleEnd': 0, 'fps': 25.0, 'startDate': None, 'description': None}, 'active': True, 'taskType': 'Generic', 'id': '415050805f1111ef93b231cb88eabe44', 'name': 'Generic', 'assignees': [], 'status': 'Not ready', 'folderId': 'c73a0ac05f1011ef93b231cb88eabe44', 'label': 'Generic', 'data': None, 'tags': []}, 'task': 'Generic'}
```
I tweaked `families = ["default"]` for both and it properly went through. But I'm not sure this is the proper fix.

robin-ynput (changes requested):

client/ayon_resolve/plugins/publish/extract_editorial_package.py
jakubjezek001: Thanks @robin-ynput, here is the commit with change 61012e27bcce5a91ea748d77856bb62c09b69637

client/ayon_resolve/plugins/publish/extract_editorial_package.py
jakubjezek001: Hey @robin-input it seems that the publishable instance was not properly created via Creator. Can you confirm please?

jakubjezek001: I have tested the workflow and it works for me 
image

The timeline was manually created and then selected and Created/converted into publishable instance. I will update the description so `Testing steps` are included

robin-ynput (approved):
I managed to do a publish + load on multiple timelines and it worked great on Resolve 19 Windows. 
Just wanted to flag that I got issues at first, it took me a while but I figured out that's when image sequence were thrown in the mix. I give an approval here cause maybe the support is out of scope.

client/ayon_resolve/plugins/load/load_editorial_package.py
robin-ynput, line 1: Maybe out of scope of this PR, so let me know if I should create a separate issue for this.

Looks like when the `editorial_pkg` is made from a Timeline which contains img sequence the loading fail as these medias could not be sourced from the `sourceClipsPath`. Main issue is that publish works but loading ends up in a confusing `NoneType object has no attribute GetName` cause the timeline is not loaded.

jakubjezek001: I remember this happening to me with the testing data we received from the client. The issue was that the first frame from the first clip was missing for some reason. I believe this is more of an edge case since we won't be dealing with offline media sources. I fixed the issue by trimming the clip to the proper length and republishing the data. After that, everything loaded perfectly.

You're right; we should at least improve user communication in case this situation occurs.

PR #14, opened 2024-08-02
Refactor timing variables check for None values

jakubjezek001:
Adjusted the check for None values in timing variables to ensure all required parameters are present before proceeding with calculations.

Testing steps:
- open resolve at at any timeline
- make sure you are having activated empty track
- open loader and load any clip which was previously published from editorial on particular clipIN/OUT position (or just set some frames in shot folder attributes)


BigRoy (approved):
Code looks good to me. :) Can't test myself now.

iLLiCiTiT (approved):

PR #3, opened 2024-07-02
Resolve: update to new publisher

iLLiCiTiT:
PR moved from https://github.com/ynput/ayon-core/pull/157

> [!NOTE] 
> Description is work in progress

## Changelog Description
Resolve with adapted latest publisher tool.

## Additional info
Adaptation of https://github.com/ynput/OpenPype/pull/5709 (thanks @bigroy for the kick-off!)

# TODO:

- [x] - @BigRoy reported that there is auto timeline created which should not 
- [ ] - @BigRoy suggested warning message if there is not anything to be used for creation
- [ ] - @BigRoy suggested to add plugin info text under `?`
- [ ] - Spliting product types into:
        	- **primary**: `shot` - instance with hierarchy, future context data and attributes placeholder
            - **secondary**: `plate`, `audio`, `review` - instance with media
- [ ] - Hierarchy and context data should be linked from Shot instance for all secondary product types Traypublisher (TP)
- [ ] - Each secondary product type which will be created from a clip should have own creator class - like in TP
- [x] - OTIO should be using native format rather then our implementation - this way we will be ready for Tony's (@tweak-wtf) retiming.

## Testing notes:
To be created

Issue #2, opened 2024-06-27
Integration fails on publishing with recent OpenTimelineIO versions (0.16.0+)


BigRoy:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior:

As can be seen in [this community forum post](https://community.ynput.io/t/ayon-monthly-beginner-s-topic-june-24/1507/22) the Resolve publishing logic fails on Precollect Instances:

![image](https://github.com/ynput/ayon-resolve/assets/2439881/1b81dbb7-1ade-4db2-a78f-1bdde13ad24d)

### Expected Behavior:

Recent `OpenTimelineIO` versions should work fine.

But, for now we could also update documentation that it requires OpenTimelineIO `0.15.0`, so along the lines of:
```
pip install OpenTimelineIO==0.15.0
```

### Version

1.0.0

### What platform you are running on?

Windows

### Steps To Reproduce:

1. Use e.g. Python 3.9 (or other Py versions like 3.10) with recent OpenTimelineIO, e.g. 0.16.0 or 0.17.0
2. Run Resolve
3. Create publishable clip
4. Publish

### Are there any labels you wish to add?

- [X] I have added the relevant labels to the bug report.

### Relevant log output:

```shell
Traceback (most recent call last):
  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2403061937_windows.zip\dependencies\pyblish\plugin.py", line 527, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-core\server_addon\resolve\client\ayon_resolve\plugins\publish\precollect_instances.py", line 114, in process
  File "E:\dev\ayon-core\server_addon\resolve\client\ayon_resolve\api\lib.py", line 912, in get_otio_clip_instance_data
    for otio_clip in otio_timeline.each_clip():
AttributeError: 'opentimelineio._otio.Timeline' object has no attribute 'each_clip'
```

BigRoy:
This is a part of the `help(timeline)` using `OpenTimelineIO==0.16.0`:
```
class Timeline(SerializableObjectWithMetadata)
 |  Method resolution order:
 |      Timeline
 |      SerializableObjectWithMetadata
 |      SerializableObject
 |      pybind11_builtins.pybind11_object
 |      builtins.object
 |  
 |  Methods defined here:
 |  
 |  __init__(...)
[...]

PR #15, opened 2024-08-28
AY-979 Finalize updating Resolve to new publisher

robin-ynput:
## Changelog Description

Finishes up with Resolve switch to new Publisher.
Finishes up the features that were already implemented in https://github.com/ynput/ayon-core/pull/157

- [x] - @BigRoy suggested warning message if there is not anything to be used for creation
- [x] - @BigRoy flagged that when clicking Create for Create Publishable Clip but there's no valid clips in timeline nothing happens
- [x] - @BigRoy flagged that previous creators was creating instances with invalid folder path(s) 
- [x] - @BigRoy flagged issues when deleting instances in the creator
- [x] - Spliting product types into:
        	- **primary**: `shot` - instance with hierarchy, future context data and attributes placeholder
            - **secondary**: `plate`, `audio`, `review` - instance with media
- [x] - Hierarchy and context data should be linked from Shot instance for all secondary product types Traypublisher (TP)
- [x] - Each secondary product type which will be created from a clip should have own creator class - like in TP
- [x] - When possible OTIO export should be done via native function rather then our implementation
- [x] - OTIO support for `opentimelineio >= 0.16.0`
- [x] - Ensure that audio cannot be exported from non-audio timeline(s)
- [x] - Test out with Resolve 19
- [x] - Help description in "?" button
- [x] - Fix review product not published when review_track contains a space
- [x] - Update client/README documentation

## Testing notes:

In Resolve:
1. Create a timeline with clips associated to a media
2. (Optional) Make a selection coloring some clips in *Chocolate* color
3. From `Workspace -> Script -> AYON Menu`
4. Create and publish product(s) of type shot, audio, plate and review from there

Tested locally on Windows with Resolve `19.0, 18.5` and OpenTimelineIO `0.17.0, 0.16.0`.


client/ayon_resolve/plugins/publish/precollect_plates.py
```diff
@@ -0,0 +1,20 @@
...
+
+    def process(self, instance):
+        """
+        Args:
+            instance (pyblish.Instance): The shot instance to update.
+        """
+        # Temporary disable no-representation failure.
+        # TODO not sure what should happen for the plate.
+        instance.data["folderPath"] = instance.data.pop("hierarchy_path")
+        instance.data["integrate"] = False
```
robin-ynput, line 20: Plate generation did not work for me:
plate_norepresentation

For now I've disabled its integration in Resolve as I believe this is something common between Hiero, Resolve and Flame that would need to be addressed in `ayon_core`. Please let me know if I'm wrong.

client/ayon_resolve/plugins/publish/precollect_shots.py
```diff
@@ -0,0 +1,86 @@
...
+        """
+        TODO: explain
+        resolve:
+        https://github.com/ynput/ayon-core/blob/6a07de6eb904c139f6d346fd6f2a7d5042274c71/client/ayon_core/plugins/publish/collect_hierarchy.py#L65
+
+        traypublisher:
+        https://github.com/ynput/ayon-traypublisher/blob/develop/client/ayon_traypublisher/plugins/publish/collect_shot_instances.py#L188
+        """
+        instance.data["folderPath"] = instance.data.pop("hierarchy_path")
+        instance.data["integrate"] = False  # no representation for shot
```
robin-ynput, line 25: This is more of a work-around for now, but probably something worth to discuss with @jakubjezek001 and  @iLLiCiTiT.
* `folderPath` as to be set to an existing folder for the Creator to detect the instance has valid context
* `folderPath` has to be the hierarchy of the shot to create for "Collect Hierarchy" plugin to properly [compute the final path](  https://github.com/ynput/ayon-core/blob/6a07de6eb904c139f6d346fd6f2a7d5042274c71/client/ayon_core/plugins/publish/collect_hierarchy.py#L65)

As a work-around this code:
1. Create a instance with `folderPath` containing the selected parent context folder at Creation to produce a valid instance
2. Immediately replace `folderPath` with final hierarchy to create in early collect plugins

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -1,23 +1,136 @@
...
-    product_type = "clip"
+    product_type = "editorial"
     icon = "film"
     defaults = ["Main"]
 
-    create_allow_context_change = False
+#    create_allow_context_change = False  
+# TODO: explain consequence on folderPath
+# https://github.com/ynput/ayon-core/blob/6a07de6eb904c139f6d346fd6f2a7d5042274c71/client/ayon_core/tools/publisher/widgets/create_widget.py#L732
+
```
robin-ynput, line 133: If `create_allow_context_change = False` then the create instance_data does not contain the selected `folderPath`.
See: https://github.com/ynput/ayon-core/blob/6a07de6eb904c139f6d346fd6f2a7d5042274c71/client/ayon_core/tools/publisher/widgets/create_widget.py#L732

This looks like a bug to me, would you like me to create an issue for that ?
Is there a better work-around than simply making it editable ?

client/ayon_resolve/plugins/create/create_workfile.py
```diff
@@ -17,8 +22,61 @@ class CreateWorkfile(AutoCreator):
...
+        # Store as marker note data
+        note = json.dumps(data)
+
+        timeline.AddMarker(
+            timeline.GetStartFrame(),
+            constants.ayon_marker_color,
+            constants.ayon_marker_name,
+            note,
+            constants.ayon_marker_duration
+        )
```
robin-ynput, line 56: @jakubjezek001 Unfortunately it does not look like Resolve allow to register custom metadata at Timeline/Track level (plus Metadata gets totally ignored by the native OTIO exporter right now).

Best solution I could think of was to store the serialized workfile instance along timeline resolution metadata as a marker on the first frame of the timeline itself. 
image
 
 Is that a good idea ? Do you have a better idea/recommendation ?

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -48,42 +161,22 @@ def header_label(text):
...
+                    tooltip="Export Plate output(s)",
+                    default=False),
+            BoolDef("export_review",
+                    label="Review",
+                    tooltip="Export Review output(s)",
+                    default=False),
+            BoolDef("export_audio",
+                    label="Audio",
+                    tooltip="Export Audio output(s)",
+                    default=False),
```
jakubjezek001, line 179: I suggest we remove the section related to Traypublisher editorial workflow. In Hiero/Resolve/Flame, we can address this by using a combination of `productType`, `reviewTrack`, and `audio` fields. Currently, we need to decide how to handle the `take` type in the `productType` enum field, as it's not yet implemented.

client/ayon_resolve/plugins/publish/precollect_plates.py
jakubjezek001: To integrate clip resources effectively in the plugin `ayon-core\client\ayon_core\plugins\publish\collect_otio_subset_resources.py`, we need to include the `clip` family in the `instance.data["families"]`. This will ensure that the conversion process accounts for the clip resources as representations that can be integrated smoothly.

jakubjezek001, line 20:
```suggestion

```

BigRoy: @jakubjezek001 Can you explain this suggestion?

jakubjezek001: We also have to include following instance data key for every instance which is having nonexistent hierarchy at the moment of publishing. 

`instance.data["newHierarchyIntegration"] = True`

client/ayon_resolve/plugins/publish/precollect_shots.py
jakubjezek001: Once `newHierarchyIntegration` will be included you can fill `folderPath` value with Ayon marker value `folder_path`

client/ayon_resolve/plugins/publish/precollect_plates.py
robin-ynput: Yep I had the "newHierarchyIntegration" set but I can confirm adding the "clip" family to plate now gives it a representation.

robin-ynput: addressed !

client/ayon_resolve/plugins/publish/precollect_review.py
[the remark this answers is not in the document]
robin-ynput: sure thing

client/ayon_resolve/plugins/publish/precollect_shots.py
robin-ynput: I guess there is something I'm doing wrong then.
I have set `newHierarchyIntegration = True` to each of the created instances.

Here is what I get if I set `folderPath` to Ayon marker value `folder_path` when creating the instance:
invalid_context

This is what I get with my hack
(set `folderPath` to parent context when creating it then adjust it at precollect time):
valid_context

client/ayon_resolve/plugins/create/create_shot_clip.py
robin-ynput: @jakubjezek001 sounds good to me ! Would you mind to confirm the expected UX please ?
Just to be sure I got it right. (For now I replicated the exact same behavior then current publisher.)

image

* I'm not sure what **Use Review Track** is needed for actually and couldn't really figure out from the code
* **Renaming** seems to be applied only to the shot name but not the resulting plate product(s) name which remain `{product_type}{variant}` (e.g. plate_VideoTrack)
* When creating a timeline from movie clips, **Include Audio** did not seem to have any effect. 
_So I've implemented a new plugin to create "audio" product from the clip source range (no handles) but that might not be what's needed._

BigRoy (commented):
Testing with Resolve 19.0B build 25 (beta release) on Windows 10.

### Publisher UI errors on empty project

Opening the publisher UI in an empty project gives me:

image

```
Creator: Workfile (io.ayon.creators.resolve.workfile)

Error: 'NoneType' object has no attribute 'GetMarkers'

  File "E:\dev\ayon-core\client\ayon_core\pipeline\create\context.py", line 1009, in reset_instances
    creator.collect_instances()
  File "E:\dev\ayon-resolve\client\ayon_resolve\plugins\create\create_workfile.py", line 120, in collect_instances
    data = self._loads_data_from_marker()
  File "E:\dev\ayon-resolve\client\ayon_resolve\plugins\create\create_workfile.py", line 72, in _loads_data_from_marker
    _, marker_info = self._get_timeline_marker()
  File "E:\dev\ayon-resolve\client\ayon_resolve\plugins\create\create_workfile.py", line 61, in _get_timeline_marker
    for idx, marker_info in timeline.GetMarkers().items():
AttributeError: 'NoneType' object has no attribute 'GetMarkers'
```

---

### Do not store data into the workfile on just opening publisher UI

Opening the creator or publisher UI _directly_ stores metadata into the scene. This should instead happen on the first `save` of the publisher UI. As such, I'd say data should persist into the file preferably only on `update_instances()` if we can instead of directly creating on `collect_instances()`.

That way just opening the UI doesn't suddenly 'pollute' new stuff into your workfile.

---

### Allow more clip color filters using enum 

(could be follow up PR, but may be trivial to implement)

I wonder if it would make sense to change this "Create" toggle from 'Use clips of Chocolate color' to an ENUM that instead is:

- Filter to clip colors: `["All clips", "Red", "Blue", "White"]`

So that basically we're not enforced to that chocolate color.

---

### Multiple publish instances

There are likely reasons for it, but I'm personally a bit confused about the double instances visible in the UI. With one clip I get:

image

Why do we need a separate "Editorial Plate" and "Editorial Shot" instance? Can't those be one publish instance? Also, what happens if I disable in the UI only editorial shot for example?

---

### Older OTIO

I got this error:
```
  File "E:\dev\ayon-resolve\client\ayon_resolve\otio\utils.py", line 106, in get_marker_from_clip_index
    all_clips = otio_timeline.find_clips()
AttributeError: 'opentimelineio._otio.Timeline' object has no attribute 'find_clips'

During handling of the above exception, another exception occurred:

  File "C:\Users\User\AppData\Local\Ynput\AYON\dependency_packages\ayon_2406251801_windows.zip\dependencies\pyblish\plugin.py", line 528, in __explicit_process
    runner(*args)
  File "E:\dev\ayon-resolve\client\ayon_resolve\plugins\publish\precollect_plates.py", line 26, in process
    otio_clip, _ = utils.get_marker_from_clip_index(
  File "E:\dev\ayon-resolve\client\ayon_resolve\otio\utils.py", line 108, in get_marker_from_clip_index
    all_clips = otio_timeline.each_clips()
AttributeError: 'opentimelineio._otio.Timeline' object has no attribute 'each_clips'
```
But I suspect that's due to using an older OTIO library. I believe historically there were reasons to stick with the older version in the Resolve integrations because it didn't work with the newer ones.

We should EITHER make it super clear that with this `ayon-resolve` release you need a newer OTIO. Or make this backwards compatible and work with the older?

I was running with `OpenTimelineIO 0.15.0`
When updating to `0.17.0` the error was gone.

---

Not sure if relevant, but I was also getting this warning in publisher UI:

```
Could not find root path for remapping "ynts_char_hero_animation_v001". This may cause issues on farm.
```

image

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -223,7 +339,25 @@ def create(self, subset_name, instance_data, pre_create_data):
                 "Go into a timeline and then reset the publisher."
             )
 
-        self.log.debug(f"Selected: {self.selected}")
+        if not self.selected:
+            if pre_create_data.get("use_selection", False):
+                raise CreatorError(
+                    "No Chocolate-colored clips found from "
+                    "timeline.\n\n Try changing clip(s) color "
```
BigRoy, line 346:
```suggestion
                    "timeline.\n\nTry changing clip(s) color "
```

Fix space on second line.
image

After:
image

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -1,23 +1,142 @@
...
+
+        Args:
+            instance(List[CreatedInstance]): Instance objects which should be
+                removed.
+        """
+        for instance in instances:
+            track_item = instance.transient_data["track_item"]
+            tag_data = lib.get_timeline_item_ayon_tag(track_item)
+            instances_data = tag_data.get(_CONTENT_ID, {})
+            _ = instances_data.pop(self.identifier, None)
```
BigRoy, line 73:
```suggestion
            instances_data.pop(self.identifier, None)
```

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -240,95 +374,105 @@ def create(self, subset_name, instance_data, pre_create_data):
         # create media bin for compound clips (trackItems)
         media_pool_folder = create_bin(self.timeline.GetName())
 
+        # detecte enabled creators for review, plate and audio
```
BigRoy, line 377:
```suggestion
        # detect enabled creators for review, plate and audio
```

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -240,95 +374,105 @@ def create(self, subset_name, instance_data, pre_create_data):
...
-            update_list(List[UpdateData]): Gets list of tuples. Each item
-                contain changed instance and it's changes.
-        """
-        for created_inst, _changes in update_list:
-            track_item = created_inst.transient_data["track_item"]
-            data = created_inst.data_to_store()
-            self.log.info(f"Storing data: {data}")
-
-            lib.imprint(track_item, data)
+        """Never called, update is handled via _ResolveInstanceCreator."""
```
BigRoy, line 474: Handled by the method definition on the parent class? In what way? Would that mean you just wouldn't re-implement it here instead? or?

client/ayon_resolve/plugins/create/create_workfile.py
BigRoy: In all hosts as far as I know the "Workfile" represents the full scene file of the host application. In this case it looks like we're considering only one timeline, and that is the _active_ timeline. Which may be odd behavior? Do we need to store this data PER timeline? Or do we need it once in the full project. If the latter, then...

There doesn't seem to be any API to set `Project Notes` otherwise that could've been handy to put a JSON string.
What if we make a dedicated "Disabled" timeline that we solely use to store the metadata.

image

client/ayon_resolve/plugins/publish/precollect_shots.py
jakubjezek001: Yeah you are quite right, the creating context is not supporting this attribute. So the hack is actually the same workaround as it is done in TP. @iLLiCiTiT question is if we could improve the workflow now so we do not need to hack this?

client/ayon_resolve/plugins/create/create_shot_clip.py
jakubjezek001: Good question for @iLLiCiTiT - related to my comment here https://github.com/ynput/ayon-resolve/pull/15#discussion_r1743207592.

client/ayon_resolve/plugins/create/create_workfile.py
jakubjezek001: I'm with @BigRoy on this. Instead of the current approach, we could create a sidecar .JSON file in the work folder alongside the .DRP files. This would serve as our go-to for reading and storing metadata. As for clip-level data, we can keep using markers like we do now.

BigRoy: Sidecar files have a lot of issues in management - I rarely feel that it's worth it going that route.

It would really be much nicer if we can find a _hacky_ way to store it _within_ the resolve project.

For what it's worth - it may be worth it for you all to make an account on Blackmagic Design forum and just respond to [this topic](https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=189685&hilit=python+database#p987221) to get some visibility on that by Resolve devs, etc. plus maybe the topic already has some other hints. (Ignore the guy who created that topic!)

And look at [this hack mentioned there](https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=189685&hilit=python+database#p991541). Could that suit our needs?

client/ayon_resolve/plugins/publish/precollect_shots.py
iLLiCiTiT: Not sure when this PR should be merged? Not sure if I could get to it in next 2 weeks.

client/ayon_resolve/plugins/create/create_shot_clip.py
iLLiCiTiT: If it is set to `False`, then it is job of create plugin to fill it.

client/ayon_resolve/plugins/create/create_workfile.py
jakubjezek001: > And look at [this hack mentioned there](https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=189685&hilit=python+database#p991541). Could that suit our needs?

Yeah! This is definitely the better way. It actually got the attention in API doc but since there were not list of available  keys and value type. Thanks for the tip @BigRoy

client/ayon_resolve/plugins/create/create_shot_clip.py
jakubjezek001: Here are some notes:

image

> I'm not sure what Use Review Track is needed for actually and couldn't really figure out from the code

From the image above you can see that additional track could be added to the timeline and that can be then pointed to the **Use Review Track** field. 

> Renaming feature seems to be applied only to the shot name but not the resulting plate product(s) name which remain {product_type}{variant} (e.g. plate_VideoTrack), should we re-label as "Rename Shot" ?

Yes, that's correct for now. It works just as you described. The product name is created by combining the Product Variant field and Product Type like this: `{product_type}{Product_variant}`. We're not using product name presets at the moment because we're still deciding how to handle tasks in the future. We might need to introduce a specific task for plate publishing, as we've seen several client requests for this in the past.

> When creating a timeline from movie clips, Include Audio did not seem to have any effect. Meaning the extracted plate got audio whatever it's checked or not.
So I've implemented a new plugin to create "audio" product from the clip source range (no handles) but that might not be what's needed.

To be honest I am not sure if it ever worked. This had been migrated from Hiero but since noone was ever reporting it it might just slip our attantion. Correct behaviour should be following:

1. Include audio toggled on
2. Additional `audioMain` product instances per clip should be created
3. Publishing process should be coverting whole timeline (included audio tracks) into OTIO and then other Core plugin should use FFMPEG and covert flattened tracks into single interemediate file which is then trimmed by the original clip timing. https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_otio_audio_tracks.py

So I am assuming that main issue here was that we were ignoring audio tracks during our custom OTIO exporter.

jakubjezek001:
> Allow more clip color filters using enum

Yeah that would be great but lets add it as enhancement issue for now please.

jakubjezek001:
> Could not find root path for remapping "ynts_char_hero_animation_v001". This may cause issues on farm.

This should not be problem, since Resolve for now is not processed over farm. But this could be fixed by adding source path to the work .DRP file.

robin-ynput:
> > Allow more clip color filters using enum
> 
> Yeah that would be great but lets add it as enhancement issue for now please.

https://github.com/ynput/ayon-resolve/issues/17

client/ayon_resolve/plugins/create/create_shot_clip.py
robin-ynput: No, sorry my explanation is probably misleading. 
Basically this `CreateShotClip` creator always delegates instance creation to "sub-creators" which are instances of `_ResolveInstanceCreator` but never hold any instances itself.
Which means that while we need `update_instances`, `remove_instances` and such defined to keep it valid creator, those will never be called from the outside. 

Operations will be handled by the "sub" creators which actually generated the instance:
* update/remove of `shot` will be made through  `ResolveShotInstanceCreator(_ResolveInstanceCreator)`
* update/remove of `plate` will be made through  `EditorialPlateInstanceCreator(_ResolveInstanceCreator)`

Hopefully this clarify a bit, let me know how I could rephase those docstring maybe.

client/ayon_resolve/plugins/create/create_shot_clip.py
robin-ynput: Thanks for the clarification.

client/ayon_resolve/plugins/create/create_shot_clip.py
robin-ynput: Awesome, thanks for confirming.

client/ayon_resolve/api/constants.py
```diff
@@ -2,6 +2,7 @@
 rename_index = 0
 rename_add = 0
 
+selected_clip_color = "Chocolate"
```
iLLiCiTiT, line 5: I know that you've added only this constant (and you followed what was here), but it would nice to use upper case variable names, for constants.

client/ayon_resolve/api/lib.py
```diff
@@ -958,8 +991,8 @@ def get_otio_clip_instance_data(otio_timeline, timeline_item_data):
     return None
 
 
-def get_otio_temp_dir(project_name, anatomy=None, timeline=None) -> str:
-    """Get otio temporary directory.
+def _get_otio_temp_file(project_name=None, anatomy=None, timeline=None) -> str:
```
iLLiCiTiT, line 994: Project name should be requirement.

client/ayon_resolve/api/constants.py
jakubjezek001: Agreed. And you can also change to uppercase the rest, please.

client/ayon_resolve/api/plugin.py
```diff
@@ -7,7 +8,8 @@
 
 from ayon_core.pipeline import (
     LoaderPlugin,
-    Creator as NewCreator
+    Creator as NewCreator,
```
iLLiCiTiT, line 11: Can this be changed to be only `Creator`?

client/ayon_resolve/api/plugin.py
```diff
@@ -454,6 +456,13 @@ def convert(self):
         else:
             self.tag_data["asset"] = self.ti_name
 
+        # AYON unique identifier
+        folder_path = "/{}/{}".format(
+            self.tag_data["hierarchy"],
+            self.tag_data["asset"],
```
iLLiCiTiT, line 462: Not sure how much work it would be, but this should not be `"asset"` (openpype term) but `"folder_name"`.

client/ayon_resolve/api/plugin.py
```diff
@@ -746,3 +745,13 @@ def _store_new_instance(self, new_instance):
 
         # Add instance to current context
         self._add_instance_to_context(new_instance)
+
+
+def get_representation_files(representation):
```
iLLiCiTiT, line 750: Where is this used?

client/ayon_resolve/api/plugin.py
```diff
@@ -454,6 +456,13 @@ def convert(self):
         else:
             self.tag_data["asset"] = self.ti_name
 
+        # AYON unique identifier
+        folder_path = "/{}/{}".format(
+            self.tag_data["hierarchy"],
+            self.tag_data["asset"],
+        )
+        self.tag_data["folder_path"] = folder_path
```
iLLiCiTiT, line 464: Does this have to be set on created instance? This can be collected during publishing.

client/ayon_resolve/plugins/publish/collect_current_project.py
```diff
@@ -1,30 +1,29 @@
...
         timeline = resolve_project.GetCurrentTimeline()
-        fps = timeline.GetSetting("timelineFrameRate")
-
-        video_tracks = rapi.get_video_track_names()
 
-        # adding otio timeline to context
-        otio_timeline = davinci_export.create_otio_timeline(resolve_project)
+        video_tracks = api.get_video_track_names()
+        otio_timeline = api.export_timeline_otio(timeline)
+        current_file = resolve_project.GetName()
```
jakubjezek001, line 19:
```suggestion
        host = registered_host()
        current_file = host.get_current_workfile()
```

And also add: 
- `from ayon_core.pipeline import registered_host`

jakubjezek001: this should fix the issue from here: https://github.com/ynput/ayon-resolve/pull/15#issuecomment-2328493576

jakubjezek001:
Lets also include into this PR following improvement:

Solution form here https://github.com/ynput/ayon-resolve/pull/15/files#r1745021400 is also applicable at https://github.com/ynput/ayon-resolve/blob/feature/AY-979_finish_Resolve_new_publisher/client/ayon_resolve/plugins/publish/extract_workfile.py lines 27 to 30.

client/ayon_resolve/api/plugin.py
jakubjezek001: Yes that is residual OP key name, but we have to keep this backward compatible when it comes to loading already created markers from previouse addon versions. Some clients might need transition time for their already created project markers. Lets clearly comment those backward compatibility changes for future references. 

Here is how some comment looks `# Backwards compatible (Deprecated since 24/06/06)` but @BigRoy might have some better idea on how this should be done. 

BTW, there is a lot of backward compatibility code coments which could be already removed.

client/ayon_resolve/api/plugin.py
jakubjezek001: https://github.com/ynput/ayon-resolve/blob/feature/AY-979_finish_Resolve_new_publisher/client/ayon_resolve/plugins/load/load_clip.py#L43-L44

robin-ynput: Yep this is not new code, I realized this was missing from parent PR when testing the load features.

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -240,95 +375,105 @@ def create(self, subset_name, instance_data, pre_create_data):
...
```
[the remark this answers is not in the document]
robin-ynput: You got it

client/ayon_resolve/api/plugin.py
robin-ynput: just to flag if we rename it "folderPath" right now, it creates invalid-context instances in the publisher.

client/ayon_resolve/api/constants.py
robin-ynput: Sure thing.

client/ayon_resolve/api/plugin.py
robin-ynput: Oh ! As I flagged in the PR header I didn't put too much attention in making the marker data backward compatible. 
But if that's a thing I'll definitely ensure it is!

client/ayon_resolve/api/plugin.py
iLLiCiTiT: I think he cannot rename it to `folderPath` because publishe UI would complain that it does not exist.

client/ayon_resolve/api/plugin.py
```diff
@@ -746,3 +745,13 @@ def _store_new_instance(self, new_instance):
 
         # Add instance to current context
         self._add_instance_to_context(new_instance)
+
+
+def get_representation_files(representation):
+    anatomy = Anatomy()
```
iLLiCiTiT, line 751:
```suggestion
def get_representation_files(project_name, representation):
    anatomy = Anatomy(project_name)
```

client/ayon_resolve/api/plugin.py
```diff
@@ -746,3 +745,13 @@ def _store_new_instance(self, new_instance):
...
+    files = []
+    for file_data in representation["files"]:
+        path = anatomy.fill_root(file_data["path"])
+        files.append(path)
+    return files
```
iLLiCiTiT, line 756:
```suggestion
    return [
        anatomy.fill_root(file_data["path"])
        for file_data in representation["files"]
    ]
```

client/ayon_resolve/api/plugin.py
robin-ynput: Done !

client/ayon_resolve/api/lib.py
robin-ynput: Updated !

client/ayon_resolve/plugins/publish/collect_current_project.py
robin-ynput: this is updated !

client/ayon_resolve/api/plugin.py
robin-ynput: This is adjusted, thanks for the feedback.

client/ayon_resolve/api/plugin.py
robin-ynput: reported !

client/ayon_resolve/api/constants.py
```diff
@@ -1,19 +1,19 @@
 # Ayon sequential rename variables
-rename_index = 0
-rename_add = 0
+#rename_index = 0
+#rename_add = 0
```
BigRoy, line 3: Commented lines should start with a space after the `#` according to PEP style guide... (yes, very nitpicky; sorry!)
```suggestion
# rename_index = 0
# rename_add = 0
```


However, aside of that. Is there a reason these are now commented but still left in? Should they be removed?

client/ayon_resolve/api/lib.py
```diff
@@ -906,6 +908,32 @@ def _convert_resolve_list_type(resolve_list):
...
+    media_pool_item = clip_item.GetMediaPoolItem()
+    clip_properties = media_pool_item.GetClipProperty()
+
+    try:
+        width, height = clip_properties["Resolution"].split("x")
+    except (KeyError, ValueError):
+        width = height = None
+
+    try:
+        pixel_aspect = int(clip_properties["PAR"]) # Pixel Aspect Resolution
```
BigRoy, line 930: PEP 8: E261 at least two spaces before inline comment
```suggestion
        pixel_aspect = int(clip_properties["PAR"])  # Pixel Aspect Resolution
```

client/ayon_resolve/api/lib.py
```diff
@@ -906,6 +908,32 @@ def _convert_resolve_list_type(resolve_list):
...
+    except(KeyError, ValueError):
```
BigRoy, line 931: When would this error?

client/ayon_resolve/api/lib.py
BigRoy: Also, what's the value of "PAR"? Because, isn't pixel aspect ratio usually a float?

client/ayon_resolve/api/plugin.py
```diff
@@ -148,7 +150,7 @@ def load(self, files):
         # create mediaItem in active project bin
         # create clip media
         media_pool_item = lib.create_media_pool_item(
-            files,
+            files[0],
```
BigRoy, line 153: So - what does this fix? It seems like an odd change to be related to "the new publisher" since it's a change for the loader?

client/ayon_resolve/api/plugin.py
```diff
@@ -746,3 +745,20 @@ def _store_new_instance(self, new_instance):
...
+
+
+def get_representation_files(project_name, representation):
+    """
+    Args:
+        project_name (str): The name of the project.
+        representation (dict): The representation to inspect.
+
+    Returns:
+        list. The files associated to the representation.
```
BigRoy, line 757:
```suggestion
        list: The files associated to the representation.
```

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -1,23 +1,143 @@
...
 from ayon_core.pipeline.create import CreatorError, CreatedInstance
 from ayon_core.lib import BoolDef, EnumDef, TextDef, UILabelDef, NumberDef
 
 
+# Used as a key by the creators in order to 
+# retrieve the instances data into clip markers.
+_CONTENT_ID = "resolve_sub_products"
+
+
+class _ResolveInstanceCreator(plugin.HiddenResolvePublishCreator):
```
BigRoy, line 18: @iLLiCiTiT should this "base class" be taken out of this file so the Plug-in discovery system does not even find it and starts thinking it's a creator it needs to "process" in one way or another being just bloat to the plug-in registry at that point?

client/ayon_resolve/plugins/create/create_workfile.py
```diff
@@ -63,8 +102,12 @@ def create(self, options=None):
...
-        # TODO: Implement
-        #   This needs to be implemented to allow persisting any instance
-        #   data on resets. We'll need to decide where to store workfile
-        #   instance data reliably. Likely metadata on the *current project*?
-        pass
+        """Store changes in project metadata so they can be recollected.
+
+        Args:
+            update_list(List[UpdateData]): Gets list of tuples. Each item
+                contain changed instance and it's changes.
```
BigRoy, line 109:
```suggestion
                contain changed instance and its changes.
```

client/ayon_resolve/plugins/create/create_workfile.py
```diff
@@ -52,9 +82,18 @@ def collect_instances(self):
...
+        if not data:
+            self.log.info("Auto-creating workfile instance...")
+            data = self._create_new_instance()
+
         current_instance = CreatedInstance(
-            self.product_type, product_name, data, self)
+            self.product_type, data["productName"], data, self)
         self._add_instance_to_context(current_instance)
 
     def create(self, options=None):
```
BigRoy, line 99: I vaguely recall @iLLiCiTiT you mentioning to me that there was _still_ a reason to implement `create` for `AutoCreator` instead of doing the 'creation' inside `collect_instances()` but I can't remember what it was. I just remember fighting you about it haha, but then subsiding because you had an explanation for it needing it.

client/ayon_resolve/plugins/publish/collect_plates.py
```diff
@@ -0,0 +1,32 @@
...
+    label = "Collect Plate"
+    hosts = ["resolve"]
+    families = ["plate"]
+
+    def process(self, instance):
+        """
+        Args:
+            instance (pyblish.Instance): The shot instance to update.
+        """
+        instance.data["folderPath"] = instance.data["folder_path"]
```
BigRoy, line 19: Did we decide on keeping this `folder_path` instead of e.g. `targetFolderPath`?

client/ayon_resolve/api/plugin.py
robin-ynput: You're right, this is totally not related to the "new publisher" thing. 
Testing stuff on my side, I realized "loading as a clip" feature was broken on my branch, and this was my quick fix to get it sorted.

Thanks to your comment I digged into it a bit more. And it comes from this:
https://github.com/ynput/ayon-resolve/commit/d3ba10caf1a3b3b3990bf00b39772824bad64a3a?diff=unified&w=0#diff-bae424ef949b94556f4ade49faccf3ad7d73f3c16135597bf20a081f9ebaa711R206
`lib.create_media_pool_item` has changed to accept a file and not a list anymore. So I've replicated the previous logic one step above so the overall load logic does not change.

client/ayon_resolve/api/lib.py
robin-ynput: I looked into that a bit more, it's not quite clear what Resolve is doing. It seems to store PAR as a string.
image

So I've come ahead with this approach:
https://github.com/ynput/ayon-resolve/pull/15/files#diff-634f1d18d81de9d69a1b8517faa8142e794ee030651aee279a5982d806d20269R20
(btw this helped me a lot https://www.unravel.com.au/aspect-ratio-cheat-sheet)

I kept a default to `1.0` in case of a weird behavior from Resolve.

client/ayon_resolve/api/plugin.py
robin-ynput: Just pushed some new commits to ensure these changes remains compatible and can load markers generated from `OpenPypeData`.

client/ayon_resolve/api/plugin.py
robin-ynput: OK, so I think I've this:

* I temporarily stored what used to be set in `folder_path` into a temporary `target_folder_path` entry, so I do not overwrite `folderPath` which contains the value needed to keep the instance valid but store it in the creator_attributes instead. So `target_folder_path` will not be part of the instance.
* Then, as per @jakubjezek001 recommendation, I've implemented the replication via `editorialSharedData` into the shot collector so that `folderPath` is properly set for each product upon collection time (same logic than traypublisher).

image

client/ayon_resolve/plugins/publish/collect_plates.py
robin-ynput: This code is no more relevant, hopefully this shed some light on this:
https://github.com/ynput/ayon-resolve/pull/15#discussion_r1746214679

client/ayon_resolve/api/lib.py
```diff
@@ -906,6 +908,34 @@ def _convert_resolve_list_type(resolve_list):
...
+    try:
+        width, height = clip_properties["Resolution"].split("x")
+    except (KeyError, ValueError):
+        width = height = None
+
+    try:
+        clip_par = clip_properties["PAR"]  # Pixel Aspect Resolution
+        pixel_aspect = constants.PAR_VALUES[clip_par]
+
+    except(KeyError, ValueError): # Unknown or undetected PAR
```
BigRoy, line 933:
```suggestion
    except (KeyError, ValueError):  # Unknown or undetected PAR
```

client/ayon_resolve/plugins/create/create_shot_clip.py
iLLiCiTiT: It can be. But I have nothing agains keeping it here if this is the only file using it.

client/ayon_resolve/plugins/create/create_workfile.py
iLLiCiTiT: I remember there was a reason, now I don't remember (it's 2 years). I would say as simple as `collect_instances` should only collect existing instances (unless there is some backwards compatibility handling), `create` should create them and store the metadata to file.

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -240,95 +382,213 @@ def create(self, subset_name, instance_data, pre_create_data):
...
+                # Plate, Audio
+                # insert parent instance data to allow
+                # metadata recollection as publish time.
+                else:
+                    parenting_data = clip_instances[shot_creator_id]
+                    sub_instance_data.update({
+                        "parent_instance_id": parenting_data["instance_id"],
+                        "creator_attributes": {
+                            "folderPath": shot_folder_path,
+                            "parent_instance": parenting_data["label"],
```
jakubjezek001, line 463: Lets convert this into `TextDef` with nice label. But let's also make sure this field in non editable by adding input attribute `disabled=True`. For reference look into `ayon-core\client\ayon_core\lib\attribute_definitions.py`

jakubjezek001: Lets actually do the same for the linked child product instances:
image

I really like how TP is having this done via shared clip base class and the TextDef is added to the class methon `def get_instance_attr_defs`. Here is some reference: [ayon_traypublisher/plugins/create/create_editorial.py#L90-L119](https://vscode.dev/github/ynput/ayon-traypublisher/blob/develop/client/ayon_traypublisher/plugins/create/create_editorial.py#L90-L119)

jakubjezek001: linking here the TP pr for above suggestions https://github.com/ynput/ayon-traypublisher/pull/12

client/ayon_resolve/plugins/create/create_workfile.py
robin-ynput: No problem. This is adjusted accordingly.

client/ayon_resolve/plugins/create/create_shot_clip.py
robin-ynput: Sounds good to me, thanks for the links. Here is what I obtain :

For Plate / Audio:
image
Is the TP "Review" attribute also needed here ? again apologies but Im not entirely sure about this one.

For Shots:
image

I've left out the clipIn, clipDuration, clipOut.. read only attributes This way there is no need to query/compute all of this information but maybe that's wrong.. 
My thought was that unlike an EDL import you can visually assess your clips directly from the timeline. 
Is that all right for you ?

jakubjezek001: > I've left out the clipIn, clipDuration, clipOut.. read only attributes This way there is no need to query/compute all of this information but maybe that's wrong..

I believe a user may want to compare and review the data in AYON to ensure its accuracy. This could save us processing time, but it might also increase the time users spend on validation, and so do we during development. I suggest we add clip and frame attributes in a read-only format to streamline this process.

BigRoy: > I've left out the clipIn, clipDuration, clipOut.. read only attributes This way there is no need to query/compute all of this information but maybe that's wrong..

If we just had "Instance" **actions** in the Publisher UI we could have a "Select clip" action. 🎉

jakubjezek001: > Is the TP "Review" attribute also needed here ?

Absolutely, that's a valid point. In Resolve or Hiero, we'll have to approach things differently compared to TP. TP has a single track where plates can be flagged for review. For Resolve and Hiero, we should introduce an editable dropdown field on the hero track clips. The dropdown options should come from the names of the timeline tracks. This value is initially set during the creation phase, but users may need to adjust it later if they forget.

Additionally, we'll need to implement a new toggle called "Hero Clip Enabled" (or something along those lines). This feature is crucial for managing vertically aligned clips and is another setting users might need to modify at a later stage.

client/ayon_resolve/plugins/publish/extract_workfile.py
```diff
@@ -16,29 +18,22 @@ class ExtractWorkfile(publish.Extractor):
...
-        name = instance.data["name"]
         project = instance.context.data["activeProject"]
-        staging_dir = self.staging_dir(instance)
 
-        ext = ".drp"
-        drp_file_name = name + ext
-        drp_file_path = os.path.normpath(
-            os.path.join(staging_dir, drp_file_name))
+        host = registered_host()        
+        drp_file_path = host.get_current_workfile()
```
jakubjezek001, line 24: We could also just get the already collected path from here https://github.com/ynput/ayon-resolve/pull/15/files#diff-d7c7c39f26e05a52a09cc2380f7dbe5c82a160e83ed1118b6fb7cca9925666c7R30

jakubjezek001 (changes requested):
There are some comments and also the publishing had failed on missing `files` in plate related representation. Following is the publishing report, you can use 

[publish-report-240910-10-19.json](https://github.com/user-attachments/files/16941802/publish-report-240910-10-19.json)


`.\tools\manage.ps1 run publish-report-viewer --use-dev` and drop the json into it to see more details.

jakubjezek001:
Hey @robin-ynput I had changed base target branch to Develop and closed the original #3 since all changes were already included here.

client/ayon_resolve/plugins/publish/extract_workfile.py
robin-ynput: Good call, thanks for that.

client/ayon_resolve/plugins/create/create_shot_clip.py
robin-ynput: Here is what I got for Shot:
image

for plate:
image

(note that those toggle have currently *no affect* on the plate creation, I think I'd need a meeting or a call with you @jakubjezek001 if more work is expected for this PR about the review product)

and audio product:
image


Does this work for you guys ?

robin-ynput:
> There are some comments and also the publishing had failed on missing `files` in plate related representation. Following is the publishing report, you can use ayon-launcher in following way to view it.
> 
> [publish-report-240910-10-19.json](https://github.com/user-attachments/files/16941802/publish-report-240910-10-19.json)
> 
> `.\tools\manage.ps1 run publish-report-viewer --use-dev` and drop the json into it.

I've opened a separate issue for that, investigating it aside. I might be wrong but I think it actually comes from a wrong retimed computation by `ayon_core` from a valid OTIO data.

client/ayon_resolve/plugins/publish/precollect_shots.py
robin-ynput: This PR is adjusted to report changes from https://github.com/ynput/ayon-core/pull/894/files

jakubjezek001 (changes requested):
I tested the latest changes with the latest dev on [this GitHub pull request](https://github.com/ynput/ayon-core/pull/897), but unfortunately, it did not go well. Here, I am including the publish report and the OTIO file it created for me.

[publish-report-240919-10-57.json](https://github.com/user-attachments/files/17056322/publish-report-240919-10-57.json)

[workfileConform_otio.zip](https://github.com/user-attachments/files/17056343/workfileConform_otio.zip)

I apologize if the review changes are more related to the [https://github.com/ynput/ayon-core/pull/897](https://github.com/ynput/ayon-core/pull/897). However, you might also find that they could be related to Resolve.

robin-ynput:
> I tested the latest changes with the latest dev on [this GitHub pull request](https://github.com/ynput/ayon-core/pull/897), but unfortunately, it did not go well. Here, I am including the publish report and the OTIO file it created for me.
> 
> [publish-report-240919-10-57.json](https://github.com/user-attachments/files/17056322/publish-report-240919-10-57.json)
> 
> [workfileConform_otio.zip](https://github.com/user-attachments/files/17056343/workfileConform_otio.zip)
> 
> I apologize if the review changes are more related to the [ynput/ayon-core#897](https://github.com/ynput/ayon-core/pull/897). However, you might also find that they could be related to Resolve.

 @jakubjezek001 just to follow up on this, the fix for this issue is in 
https://github.com/ynput/ayon-core/pull/916

client/ayon_resolve/README.markdown
```diff
@@ -10,17 +10,17 @@
     ![image](https://user-images.githubusercontent.com/40640033/102792588-ffcb1c80-43a8-11eb-9c6b-bf2114ed578e.png) with installed CMake in PATH.
 -   make sure Resolve Fusion (Fusion Tab/menu/Fusion/Fusion Settings) is set to Python 3.6
     ![image](https://user-images.githubusercontent.com/40640033/102631545-280b0f00-414e-11eb-89fc-98ac268d209d.png)
--   Open OpenPype **Tray/Admin/Studio settings** > `applications/resolve/environment` and add Python3 path to `RESOLVE_PYTHON3_HOME` platform related.
+-   Open Ayon **Tray/Admin/Studio settings** > `applications/resolve/environment` and add Python3 path to `RESOLVE_PYTHON3_HOME` platform related.
```
iLLiCiTiT, line 13: It should be `AYON` instead of `Ayon`.

robin-ynput: this is adjusted, thanks.

client/ayon_resolve/api/lib.py
```diff
@@ -224,43 +223,58 @@ def create_bin(name: str,
...
 
-    # add all data in folder to media pool
-    media_pool_items = media_pool.ImportMedia(files)
+    dirname, file = os.path.split(fpath)
+    _name, ext = os.path.splitext(file)
 
-    return media_pool_items.pop() if media_pool_items else False
+    # add all data in folder to media-pool
+    media_pool_items = media_storage.AddItemListToMediaPool(
+        os.path.normpath(dirname))
```
BigRoy, line 267: Can we make this work without doing this? I'd really prefer it to NOT load ALL media from the folder to the media pool.

client/ayon_resolve/api/lib.py
```diff
@@ -953,13 +1017,96 @@ def get_otio_clip_instance_data(otio_timeline, timeline_item_data):
...
+            "Cannot use native OTIO export (%r)."
+            "Default to AYON own implementation.",
+            error
+        )
+        otio_timeline = otio_export.create_otio_timeline(
+            get_current_resolve_project(),
+            timeline=timeline
+        )
+        otio_export.write_to_file(otio_timeline, filepath)
+
```
BigRoy, line 1081:
```suggestion
```

client/ayon_resolve/api/pipeline.py
```diff
@@ -97,27 +99,33 @@ def get_workfile_extensions(self):
     def get_containers(self):
         return ls()
 
+    def get_context_data(self):
+        return {}
+
+    def update_context_data(self, data, changes):
+        pass
```
BigRoy, line 106: This is still lacking - and due to this any context option attributes in the publisher UI will not persist. Which isn't a big problem because apparently we do not have any context attributes exposed yet in the publisher UI, at least I didn't see any in resolve - but we might want to add todos here.

```suggestion
    def get_context_data(self):
        # TODO: implement to support persisting context attributes
        return {}

    def update_context_data(self, data, changes):
        # TODO: implement to support persisting context attributes
        pass
```

client/ayon_resolve/api/plugin.py
```diff
@@ -404,9 +135,15 @@ def load(self, files):
         self.active_bin = lib.create_bin(self.data["binPath"])
 
         # create mediaItem in active project bin
+
```
BigRoy, line 138:
```suggestion
```

client/ayon_resolve/plugins/publish/collect_plates.py
```diff
@@ -0,0 +1,27 @@
...
+        instance.data["families"].append("clip")
+
+        # Retrieve instance data from parent instance shot instance.
+        parent_instance_id = instance.data["parent_instance_id"]
+        edit_shared_data = instance.context.data["editorialSharedData"]
+        instance.data.update(
+            edit_shared_data[parent_instance_id]
+        )
+
+        self.log.debug(pprint.pformat(instance.data))
```
BigRoy, line 27: Can we avoid this heavy verbose logging here?
```suggestion
```

client/ayon_resolve/plugins/publish/collect_shots.py
```diff
@@ -0,0 +1,112 @@
...
+        instance.data.update(
+            {
+                "resolutionWidth": width,
+                "resolutionHeight": height,
+                "pixelAspect": pixel_aspect,
+            }
+        )
+
+        self._inject_editorial_shared_data(instance)
+        self.log.debug(pprint.pformat(instance.data))
```
BigRoy, line 112: Can we avoid this heavy verbose logging here? Or be more specific about what we want to log?
```suggestion
```

client/ayon_resolve/api/pipeline.py
BigRoy: It'd basically be an implementation similar to this https://github.com/ynput/ayon-resolve/blob/1b6e9aa353dc88d86d51ec3f41920e27ee0e441d/client/ayon_resolve/plugins/create/create_workfile.py#L24-L46 but then storing it in something that does not clash with that data.

BigRoy (approved):
This works! ✅ I did have some remarks on code and one lacking (non-urgent) feature. But other than that, amazing effort. Already so much better than the old creator stuff.

client/ayon_resolve/plugins/publish/collect_shots.py
robin-ynput: I'm gonna let @jakubjezek001 answer this one. I've basically reported it from existing plugins for consistency.

From my experience, it's bruteforce but very valuable for a developper (hence debug) especially for troubleshooting.
But if it ends up in user face that'll be quite disturbing for an artist indeed.

client/ayon_resolve/api/pipeline.py
robin-ynput: This creator class is inherited by "exposed" AYON creators implemented in `plugins/create/create_shot_clip.py` file.
I believe the code you are looking for is there:
https://github.com/ynput/ayon-resolve/blob/feature/AY-979_finish_Resolve_new_publisher/client/ayon_resolve/plugins/create/create_shot_clip.py#L120

BigRoy: No - the context data is the "global context data" of the Create context and does not store "per instance". This would still need implementation exactly here on the Host addon with the `IPublishHost` interface

See: https://ayon.ynput.io/docs/dev_publishing#required-functions-in-host-implementation

In Maya we store that in the global file info for example, see [here](https://github.com/ynput/ayon-maya/blob/ca40abe992efce70d9ef26f049a5500513ba0ea7/client/ayon_maya/api/pipeline.py#L158-L170).

jakubjezek001:
Great progress, @robin-ynput! My testing went well with the 19.5 version. I no longer have 18.5 installed, so this should be enough for now.

We're another step closer to the final state. Here are some points we found during testing:

## Future context with `has_promised_context`
image

In the image, you can see that we are still using the original implementation of editorial. The limits did not allow us to add future context into the instance context. After introducing `has_promised_content`, this no longer applies and can be refactored. The future context should be in the appropriate section, and the instance attribute **Folder path** can be removed. This applies to all highlighted instances. Only the _WorkfileConform_ instance has the correct context.

Since the `has_promised_context` attribute is in place, the Task can be left empty, and the Folder path can be set to a nonexistent value.

client/ayon_resolve/plugins/create/create_workfile.py
```diff
@@ -0,0 +1,118 @@
...
+        project_name = self.create_context.get_current_project_name()
+        folder_path = self.create_context.get_current_folder_path()
+        task_name = self.create_context.get_current_task_name()
+        host_name = self.create_context.host_name
+
+        folder_entity = ayon_api.get_folder_by_path(
+            project_name, folder_path)
+        task_entity = ayon_api.get_task_by_name(
+            project_name, folder_entity["id"], task_name
+        )
```
iLLiCiTiT, line 60:
```suggestion
        project_name = self.create_context.get_current_project_name()
        host_name = self.create_context.host_name
        folder_entity = self.get_current_folder_entity()
        task_entity = self.get_current_task_entity()
        folder_path = folder_entity["path"]
        task_name = task_entity["name"]
        product_name = self.get_product_name(
            project_name,
            folder_entity,
            task_entity,
            self.default_variant,
            host_name,
        )
```

client/ayon_resolve/api/constants.py
```diff
@@ -0,0 +1,25 @@
...
+SELECTED_CLIP_COLOR = "Chocolate"
+PUBLISH_CLIP_COLOR = "Pink"
+AYON_MARKER_WORKFLOW = True
+
+# Ayon compound clip workflow variable
+AYON_TAG_NAME = "VFX Notes"
+
+# Ayon marker workflow variables
+LEGACY_OPENPYPE_MARKER_NAME = "OpenPypeData"
+AYON_MARKER_NAME = "AyonData"
```
iLLiCiTiT, line 11:
```suggestion
AYON_MARKER_NAME = "AYONData"
```

client/ayon_resolve/api/constants.py
```diff
@@ -0,0 +1,25 @@
...
+AYON_MARKER_DURATION = 1
+AYON_MARKER_COLOR = "Mint"
+TEMP_MARKER_FRAME = None
+
+# Ayon default timeline
+AYON_TIMELINE_NAME = "AyonTimeline"
```
iLLiCiTiT, line 17:
```suggestion
AYON_TIMELINE_NAME = "AYONTimeline"
```

client/ayon_resolve/utility_scripts/ayon_startup.scriptlib
```diff
@@ -1,4 +1,4 @@
--- Run OpenPype's Python launch script for resolve
+-- Run Ayon's Python launch script for resolve
```
iLLiCiTiT, line 1:
```suggestion
-- Run AYON's Python launch script for resolve
```

client/ayon_resolve/utility_scripts/ayon_startup.scriptlib
```diff
@@ -12,7 +12,7 @@ if ayon_startup_script ~= nil then
     if file_exists(script) then
         -- We must use RunScript to ensure it runs in a separate
         -- process to Resolve itself to avoid a deadlock for
-        -- certain imports of OpenPype libraries or Qt
+        -- certain imports of Ayon libraries or Qt
```
iLLiCiTiT, line 15:
```suggestion
        -- certain imports of AYON libraries or Qt
```

client/ayon_resolve/plugins/publish/collect_shots.py
BigRoy: If this is the kind of debugging we need - I'd argue that we should instead focus on allowing such functionality inside the Publisher UI itself (similar to the Pyblish Debug Stepper https://github.com/ynput/ayon-core/pull/753) so that one could e.g. have the publisher itself set e.g. debug flags for certain plug-ins and then e.g. add a "profiled" report that e.g. reports the "changes" to the `instance.data` from that plug-in, etc.

We should really try and refrain from having each plug-in enforcing this massive data point all the time. It's slow - and on my old machine; it's **very slow**. The majority of my publish time is spent around these calls ;) (luckily these are without the `json.dumps` around it because those were even slower)

robin-ynput: That makes sense, especially if it impacts performances. I'll remove those, I guess a developer can easily bring them back if needed while troubleshooting.

client/ayon_resolve/api/pipeline.py
robin-ynput: OK, I'll report the TODOs then.

client/ayon_resolve/api/lib.py
robin-ynput: Yeah, I've restore current implementation. I think this changes were on initial draft PR but they do not seem relevant with this PR. I could confirm it all work successfully.

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -1,245 +1,490 @@
...
+            ),
+
+            # publishSettings
+            UILabelDef(
+                label=header_label("Clip Publish Settings")
+            ),
+            EnumDef(
+                "clip_variant",
+                label="Product Variant",
+                tooltip="Chose variant which will be then used for "
```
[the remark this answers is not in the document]
robin-ynput: thanks for that

robin-ynput:
@jakubjezek001 I've addressed your feedback regarding the `has_promised_context`. Thanks for that.
adjusted_feedback_resolve

Regarding your question about handles workflow.
So, I have never been an editor myself and always was on the tooling side, but I don't recall any dynamic/min handles workflow. 
As far as I can remember, Editors were creating shots in Ftrack/Shotgrid with handles defined from client specs document and keep it consistent per shot. 
So the logic we have in `extract_otio_review` (extra from media if available or generate black frames otherwise) is what I am more used to.

jakubjezek001 (approved):
All works fine as expected! Great work @robin-ynput - we can merge this but lets wait also for the merge of https://github.com/ynput/ayon-core/pull/916 and ideally after it is released, so we could change the dependent core version in `ayon-resolve\package.py` under **ayon_required_addons**.

robin-ynput:
TODO:
- [ ] Deploy and release https://github.com/ynput/ayon-core/pull/897
- [ ] Adjust `package.py` requirement accordingly

jakubjezek001:
It looks like the linked issue might have been added by mistake. Is it correct? After a quick look at the changes, it seems that the suggested changes from issue #2 weren't applied.

robin-ynput:
> It looks like the linked issue might have been added by mistake. Is it correct? After a quick look at the changes, it seems that the suggested changes from issue #2 weren't applied.

This is not a mistake, I've tested all of the work on Windows with OTIO 0.17.0 and 0.16.0 it works. 
So the issue described on this issue has been fixed but with another approach than the suggested change on the ticket (wrapper function + enforce documentation so people use 0.15.0 version).

PR #36, opened 2024-10-07
Transfer editorial package to new publisher

robin-ynput:
## Changelog Description
Transfer `editorial_pkg` creator to new publisher.
This contains changes from #12. 

_Note_ this is backward-compatible with previously save scene and published packages as metadata remains store under the same key `VFX Notes` which used to be referenced as `api.lib.pype_tag_name` but now becomes `api.constants.AYON_TAG_NAME`

## Testing notes:

1. Create a timeline
2. Select this timeline
3. Create a new `editorial pkg` from `AYON Menu` using `Editorial Package` creator
4. Publish it
5. Load the published timeline using the Loader

### DIFF
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -707,6 +707,7 @@ def get_editorial_publish_data(
         folder_path (str): Folder path where editorial package is located.
         product_name (str): Editorial product name.
         version (Optional[str]): Editorial product version. Defaults to None.
+        task (Optional[str]): Associated task name. Defaults to None (no task).
 
     Returns:
         dict: Editorial publish data.
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -1,4 +1,4 @@
-import copy 
+import copy
 import re
 import uuid
 
@@ -8,7 +8,7 @@ from ayon_core.pipeline.constants import AVALON_INSTANCE_ID
 from ayon_core.pipeline import (
     LoaderPlugin,
     Creator,
-    HiddenCreator,    
+    HiddenCreator,
     Anatomy
 )
 
@@ -698,7 +698,8 @@ class ResolvePublishCreator(Creator):
 def get_editorial_publish_data(
     folder_path,
     product_name,
-    version=None
+    version=None,
+    task=None,
 ) -> dict:
     """Get editorial publish data from context.
 
@@ -722,6 +723,9 @@ def get_editorial_publish_data(
 
     if version:
         data["version"] = version
+
+    if task:
+        data["task"] = task
 
     return data
 
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -53,7 +53,7 @@ class CreateEditorialPackage(ResolveCreator):
 
         new_instance = CreatedInstance(
             self.product_type,
-            publish_data["publish"]["productName"],
+            product_name,
             publish_data,
             self,
         )
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -16,15 +16,15 @@ class CreateEditorialPackage(ResolveCreator):
     icon = "camera"
     defaults = ["Main"]
 
-    def create(self, subset_name, instance_data, pre_create_data):
+    def create(self, product_name, instance_data, pre_create_data):
         """Create a new editorial_pkg instance.
 
         Args:
-            subset_name (str): The subset name
+            product_name (str): The subset name
             instance_data (dict): The instance data.
             pre_create_data (dict): The pre_create context data.
         """
-        super().create(subset_name,
+        super().create(product_name,
                        instance_data,
                        pre_create_data)
 
@@ -39,13 +39,6 @@ class CreateEditorialPackage(ResolveCreator):
 
         publish_data = deepcopy(instance_data)
 
-        # add publish data for streamlrine publishing
-        product_name = self.get_product_name(
-            self.project_name,
-            self.create_context.get_current_folder_entity(),
-            self.create_context.get_current_task_entity(),
-            instance_data["variant"],
-        )
         publish_data["publish"] = get_editorial_publish_data(
             folder_path=instance_data["folderPath"],
             product_name=product_name,
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -20,9 +20,9 @@ class CreateEditorialPackage(ResolveCreator):
     def create(self, subset_name, instance_data, pre_create_data):
         """
         """
-        super(CreateEditorialPackage, self).create(subset_name,
-                                           instance_data,
-                                           pre_create_data)
+        super().create(subset_name,
+                       instance_data,
+                       pre_create_data)
 
         current_timeline = lib.get_current_timeline()
 
@@ -74,7 +74,7 @@ class CreateEditorialPackage(ResolveCreator):
 
             # exclude all which are not productType editorial_pkg
             if (
-                data.get("publish", {}).get("productType") != "editorial_pkg"
+                data.get("publish", {}).get("productType") != self.product_type
             ):
                 continue
 
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -11,14 +11,18 @@ class CreateEditorialPackage(ResolveCreator):
     """Create Editorial Package."""
 
     identifier = "io.ayon.creators.resolve.editorial_pkg"
-    product_name = "editorial_pkgMain"
     label = "Editorial Package"
     product_type = "editorial_pkg"
     icon = "camera"
     defaults = ["Main"]
 
     def create(self, subset_name, instance_data, pre_create_data):
-        """
+        """Create a new editorial_pkg instance.
+
+        Args:
+            subset_name (str): The subset name
+            instance_data (dict): The instance data.
+            pre_create_data (dict): The pre_create context data.
         """
         super().create(subset_name,
                        instance_data,
@@ -35,25 +39,32 @@ class CreateEditorialPackage(ResolveCreator):
 
         publish_data = deepcopy(instance_data)
 
-        # add publish data for streamline publishing
+        # add publish data for streamlrine publishing
+        product_name = self.get_product_name(
+            self.project_name,
+            self.create_context.get_current_folder_entity(),
+            self.create_context.get_current_task_entity(),
+            instance_data["variant"],
+        )
         publish_data["publish"] = get_editorial_publish_data(
             folder_path=instance_data["folderPath"],
-            product_name=self.product_name,
+            product_name=product_name,
         )
 
-        publish_data["label"] = current_timeline.GetName()
+        publish_data.update({
+            "label": current_timeline.GetName(),
+        })
         timeline_media_pool_item.SetMetadata(
             constants.AYON_TAG_NAME, json.dumps(publish_data)
         )
 
-        publish_data["media_pool_item_id"] = timeline_media_pool_item.GetUniqueId()
         new_instance = CreatedInstance(
             self.product_type,
-            self.product_name,
+            publish_data["publish"]["productName"],
             publish_data,
             self,
         )
-        new_instance.transient_data["timeline_item"] = timeline_media_pool_item
+        new_instance.transient_data["timeline_pool_item"] = timeline_media_pool_item
         self._add_instance_to_context(new_instance)
 
     def collect_instances(self):
@@ -78,15 +89,14 @@ class CreateEditorialPackage(ResolveCreator):
             ):
                 continue
 
-            data["media_pool_item_id"] = media_pool_item.GetUniqueId()
             current_instance = CreatedInstance(
                 self.product_type,
-                self.product_name,
+                data["publish"]["productName"],
                 data,
                 self
             )
 
-            current_instance.transient_data["timeline_item"] = media_pool_item            
+            current_instance.transient_data["timeline_pool_item"] = media_pool_item            
             self._add_instance_to_context(current_instance)
 
     def update_instances(self, update_list):
@@ -97,7 +107,7 @@ class CreateEditorialPackage(ResolveCreator):
                 contain changed instance and it's changes.
         """
         for created_inst, _changes in update_list:
-            timeline_media_pool_item = created_inst.transient_data["timeline_item"]
+            timeline_media_pool_item = created_inst.transient_data["timeline_pool_item"]
             timeline_media_pool_item.SetMetadata(
                 constants.AYON_TAG_NAME,
                 json.dumps(created_inst.data_to_store()),
@@ -112,7 +122,7 @@ class CreateEditorialPackage(ResolveCreator):
         """
         for instance in instances:
             self._remove_instance_from_context(instance)
-            timeline_media_pool_item = instance.transient_data["timeline_item"]
+            timeline_media_pool_item = instance.transient_data["timeline_pool_item"]
             timeline_media_pool_item.SetMetadata(
                 constants.AYON_TAG_NAME,
                 json.dumps({}),
--- a/client/ayon_resolve/plugins/create/create_editorial_package.py
+++ b/client/ayon_resolve/plugins/create/create_editorial_package.py
@@ -1,88 +1,120 @@
 import json
 from copy import deepcopy
-from ayon_core.tools.context_dialog.window import (
-    ContextDialog,
-    ContextDialogController
-)
-from ayon_core.pipeline import get_current_project_name
-from ayon_api import get_folder_by_id, get_task_by_id, get_folder_by_path
-from ayon_core.pipeline.create.legacy_create import LegacyCreator
+
+from ayon_core.pipeline.create import CreatorError, CreatedInstance
 
 from ayon_resolve.api import lib, constants
-from ayon_resolve.api.plugin import get_editorial_publish_data
+from ayon_resolve.api.plugin import ResolveCreator, get_editorial_publish_data
 
 
-class CreateEditorialPackage(LegacyCreator):
+class CreateEditorialPackage(ResolveCreator):
     """Create Editorial Package."""
 
-    name = "editorial_pkg"
+    identifier = "io.ayon.creators.resolve.editorial_pkg"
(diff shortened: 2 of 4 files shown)


client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,120 @@
...
+from ayon_resolve.api.plugin import ResolveCreator, get_editorial_publish_data
 
-class CreateEditorialPackage(LegacyCreator):
+
+class CreateEditorialPackage(ResolveCreator):
     """Create Editorial Package."""
 
-    name = "editorial_pkg"
+    identifier = "io.ayon.creators.resolve.editorial_pkg"
+    product_name = "editorial_pkgMain"
```
BigRoy, line 14: A product name should technically not be hardcoded or overridden like this - but should adhere to the anatomy templates.

In this case you wouldn't need this at all and this product name appears to be the 'default' for the default anatomy template anyway. So I'd say, remove this.

(By default it uses the `Creator.get_product_name` implementation, if you're curious)

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,120 @@
...
-    def process(self):
-        """Process the creation of the editorial package."""
-        project_name = get_current_project_name()
-        folder_path = self.data["folderPath"]
-
-        current_folder = get_folder_by_path(project_name, folder_path)
-        print(current_folder)
+    def create(self, subset_name, instance_data, pre_create_data):
+        """
+        """
```
BigRoy, line 22: Fill in docstring, or remove.

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,120 @@
...
+        super(CreateEditorialPackage, self).create(subset_name,
+                                           instance_data,
+                                           pre_create_data)
```
BigRoy, line 25: Should be fine like this?
```suggestion
        super().create(subset_name,
                       instance_data,
                       pre_create_data)
```

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,120 @@
...
+            product_name=self.product_name,
         )
 
+        publish_data["label"] = current_timeline.GetName()
         timeline_media_pool_item.SetMetadata(
-            lib.pype_tag_name, json.dumps(publish_data)
+            constants.AYON_TAG_NAME, json.dumps(publish_data)
         )
 
+        publish_data["media_pool_item_id"] = timeline_media_pool_item.GetUniqueId()
```
BigRoy, line 49: Why does it need this stored, if we're also adding the actual item as `transient_data["timeline_item"]`?

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,120 @@
...
+            except json.JSONDecodeError:
+                self.log.warning(
+                    f"Failed to parse json data from media pool item: "
+                    f"{media_pool_item.GetName()}"
+                )
+                continue
+
+            # exclude all which are not productType editorial_pkg
+            if (
+                data.get("publish", {}).get("productType") != "editorial_pkg"
```
BigRoy, line 77: Best to use the class variable here?
```suggestion
                data.get("publish", {}).get("productType") != self.product_type
```

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,120 @@
...
+            ):
+                continue
+
+            data["media_pool_item_id"] = media_pool_item.GetUniqueId()
```
BigRoy, line 81: Same question as above - why do we need to store the id?

client/ayon_resolve/plugins/publish/collect_editorial_package.py
```diff
@@ -1,88 +1,56 @@
...
     label = "Collect Editorial Package Instances"
     families = ["editorial_pkg"]
 
-    def process(self, context):
-        project_name = context.data["projectName"]
+    def process(self, instance):
+        project_name = instance.context.data["projectName"]
         self.log.info(f"project: {project_name}")
 
+        media_pool_item_id = instance.data["media_pool_item_id"]
```
BigRoy, line 19: You're better of here doing:
```suggestion
        media_pool_item = instance.data["transientData"]["timeline_item"]
```

However, the usage of `timeline_item` here is maybe confusing? Because the media pool item isn't a timeline item, is it? So we should maybe also refactor that transient data variable name?

client/ayon_resolve/plugins/publish/collect_editorial_package.py
```diff
@@ -1,88 +1,56 @@
...
+                    version = last_version + 1
 
-            if publish_data.get("taskId"):
-                task_entity = get_task_by_id(
-                    project_name=project_name,
-                    task_id=publish_data["taskId"],
-                )
-                publish_data["taskEntity"] = task_entity
-                publish_data["task"] = task_entity["name"]
+            instance.data["version"] = version
```
BigRoy, line 47: This seems redundant - an instance should always automatically pick a version that is the next version that is available?

We should be able to take this out completely?

Which means - that if this is removed, maybe this full collector becomes redundant because the media pool item is available in the transient data - and hence... it's not collecting anything of use anymore? We may just need to update any other usage of `instance.data["mediaPoolItem"` and `instance.data["item"]` for this family?

BigRoy:
[This](https://github.com/ynput/ayon-resolve/blob/0ca7b232b67e293a0ce3f3d934548a399e572af6/client/ayon_resolve/api/plugin.py#L728) should likely also be a `if version is not None` check because a version of 0 is technically a valid version?

client/ayon_resolve/plugins/create/create_editorial_package.py
robin-ynput: Thanks for this suggestion, I didn't know about this method. 
Hopefully I've used it correctly, please let me know if not.

client/ayon_resolve/plugins/publish/collect_editorial_package.py
robin-ynput: I've renamed it to `timeline_pool_item`

client/ayon_resolve/plugins/publish/collect_editorial_package.py
robin-ynput: Good call, that's something I expected @jakubjezek001 could shed some light on. 
I initially thought this could be removed because this whole logic seems trigged when provided `version` is not None and the creator does not set this up. But wanted to check with you first in case that `version` edited by some hidden mechanism..

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,130 @@
...
-        current_folder = get_folder_by_path(project_name, folder_path)
-        print(current_folder)
+        Args:
+            subset_name (str): The subset name
+            instance_data (dict): The instance data.
+            pre_create_data (dict): The pre_create context data.
+        """
+        super().create(subset_name,
+                       instance_data,
+                       pre_create_data)
```
BigRoy, line 29:
```suggestion
        Args:
            product_name (str): The subset name
            instance_data (dict): The instance data.
            pre_create_data (dict): The pre_create context data.
        """
        super().create(product_name,
                       instance_data,
                       pre_create_data)
```

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,130 @@
...
     label = "Editorial Package"
     product_type = "editorial_pkg"
     icon = "camera"
     defaults = ["Main"]
 
-    def process(self):
-        """Process the creation of the editorial package."""
-        project_name = get_current_project_name()
-        folder_path = self.data["folderPath"]
+    def create(self, subset_name, instance_data, pre_create_data):
```
BigRoy, line 19:
```suggestion
    def create(self, product_name, instance_data, pre_create_data):
```

client/ayon_resolve/plugins/create/create_editorial_package.py
BigRoy: Ah this one is on me - because I missed one bit which seeing it now... made me understand why you're doing this.
You can remove the manual calling of the method - because it's done for you. See [this comment](https://github.com/ynput/ayon-resolve/pull/36/files#r1790937622).

The `Creator.create` method's first argument is the `product_name`. Which it gets passed from `CreateContext.create` here: https://github.com/ynput/ayon-core/blob/8af66a0f783a959ca477418618775fe9081f6574/client/ayon_core/pipeline/create/context.py#L1153-L1171

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,130 @@
...
-        # add publish data for streamline publishing
+        publish_data = deepcopy(instance_data)
+
+        # add publish data for streamlrine publishing
+        product_name = self.get_product_name(
+            self.project_name,
+            self.create_context.get_current_folder_entity(),
+            self.create_context.get_current_task_entity(),
+            instance_data["variant"],
+        )
```
BigRoy, line 48:
```suggestion
```

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,130 @@
...
+            "label": current_timeline.GetName(),
+        })
         timeline_media_pool_item.SetMetadata(
-            lib.pype_tag_name, json.dumps(publish_data)
+            constants.AYON_TAG_NAME, json.dumps(publish_data)
         )
 
+        new_instance = CreatedInstance(
+            self.product_type,
+            publish_data["publish"]["productName"],
```
BigRoy, line 63: This is just `product_name` argument already defined above.

robin-ynput: Yes I understood better what you meant, make more sense and simpler code indeed. Thanks for that.

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,123 @@
...
-        """Process the creation of the editorial package."""
-        project_name = get_current_project_name()
-        folder_path = self.data["folderPath"]
+    def create(self, product_name, instance_data, pre_create_data):
+        """Create a new editorial_pkg instance.
 
-        current_folder = get_folder_by_path(project_name, folder_path)
-        print(current_folder)
+        Args:
+            product_name (str): The subset name
```
BigRoy, line 23:
```suggestion
            product_name (str): The product name
```

jakubjezek001:
> @jakubjezek001 there is one side-effect I found with this update. When created the `editorial_pkg` instance adds new metadata into the timeline media_pool clip, which eventually get published. Then, when you load the result, the newly created timeline will bring back the `editorial_pkg` metadata used for its creation and this will mean a new instance might appear out of nowhere in the creator. Technically this makes sense, but I guess it could be confusing for the artist. Should we clear the editorial_pkg data upon loading ?
> 
> (hopefully I that make sense, sorry if not)

Hey @robin-ynput, I tested the workflow and added some enhancements for loading and publishing data distribution. You can find the changes here: bfd9df5b7e93b7eb17f8d33573e7f7f9357c976c. The unified data distribution for loading and publishing is specific to this product type, as requested by the client.

jakubjezek001 (approved):
Tested and it works fine now after the enhancements.

robin-ynput:
Thanks a lot @jakubjezek001 for your enhancements. Retested everything on my side as well, works perfectly.
@BigRoy do you need more time to give an approval on this one ? I'd like to go ahead and merge it when possible.

client/ayon_resolve/plugins/publish/collect_editorial_package.py
jakubjezek001: Yeah this workflow is looking strange but it is the client preference. Use next available version if timeline is loaded from editorial package (this way we preserve the explicit chain), if timeline is created via Creator then find next available arbitrary version.

client/ayon_resolve/api/plugin.py
```diff
@@ -698,7 +698,8 @@ def remove_instances(self, instances):
 def get_editorial_publish_data(
     folder_path,
     product_name,
-    version=None
+    version=None,
+    task=None,
```
BigRoy, line 702: This is missing in the docstring

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,155 @@
...
+        for media_pool_item in lib.iter_all_media_pool_clips():
+            data = media_pool_item.GetMetadata(constants.AYON_TAG_NAME)
+            if not data:
+                continue
+
+            try:
+                data = json.loads(data)
+            except json.JSONDecodeError:
+                self.log.warning(
+                    f"Failed to parse json data from media pool item: "
```
BigRoy, line 76:
```suggestion
                    "Failed to parse json data from media pool item: "
```

client/ayon_resolve/plugins/publish/collect_editorial_package.py
```diff
@@ -1,88 +1,51 @@
...
-            instance.data.update(publish_data)
+        instance.data.update(
+            {
+                "mediaPoolItem": media_pool_item,
+                "item": media_pool_item,
+            }
+        )
 
-            self.log.info(f"Editorial Package: {instance.data}")
+        self.log.info(f"Editorial Package: {instance.data}")
```
BigRoy, line 51: Pretty sure this should be debug log since it's not artist-facing information.

```suggestion
        self.log.debug(f"Editorial Package: {instance.data}")
```

client/ayon_resolve/plugins/publish/collect_editorial_package.py
```diff
@@ -1,88 +1,51 @@
...
-                {
-                    "mediaPoolItem": media_pool_item,
-                    "item": media_pool_item,
-                }
+            # make sure last version of product is higher than current
+            # expected current version from publish data
+            folder_entity = ayon_api.get_folder_by_path(
+                project_name=project_name,
+                folder_path=instance.data["folderPath"],
+            )
```
BigRoy, line 31: Isn't this just `instance.data["folderEntity"]` ?

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,155 @@
...
+            ):
+                continue
+
+            publish_data = data["publish"]
+
+            # TODO: backward compatibility for legacy workflow instances
+            # add label into instance data in case it is missing in publish
+            # data
+            if "label" not in publish_data:
+                publish_data["label"] = media_pool_item.GetName()
```
BigRoy, line 93: Why is `label` data so important that we require it in `instance.data`? Can't we just ignore it not existing? (And maybe even more forward with not setting it now, or?)

BigRoy: So I understood from Robin this is purely cosmetic - we just want to display the timeline name in the publisher instead of the product name.

Hence, I'd just ALWAYS set this and disregard it from whatever is imprintend on the instance.

If someone renames the timeline, we want it display the new name - not whatever was once imprinted to it on the past.

client/ayon_resolve/plugins/publish/collect_editorial_package.py
```diff
@@ -1,88 +1,51 @@
...
-                data.get("publish")
-                and data["publish"].get("productType") != "editorial_pkg"
-            ):
-                continue
-
-            instance = context.create_instance(name=media_pool_item.GetName())
+        media_pool_item = instance.data["transientData"]["timeline_pool_item"]
 
-            publish_data = data["publish"]
+        # get version from publish data and rise it one up
```
BigRoy, line 21: Since this logic existed before this PR and this matches it - let's keep it in. It would be nice though @robin-ynput if here you could elaborate on what this does and why (like a small elaboration on it).

And then separate from this PR let's create an issue to track that we will want to change this logic. It's a very poor-man's solution to what should not be so specific to this particular instance/product.

If we want _that type of versioning_ we should document that clearly, define the strategy and potentially make that "just one of the ways you can version products" in AYON. Now it just seems way too hardcoded for just this resolve use case (which I still think isn't the correct use case, but that's up for debate in a separate issue.)

robin-ynput: As per our discussion, relevant comment has been added.

client/ayon_resolve/api/plugin.py
robin-ynput: Adjusted !

client/ayon_resolve/plugins/publish/collect_editorial_package.py
robin-ynput: Update, we had a call aside and agreed to leave this one as-is for now. An explanatory comment has been added in the plugin tho.

client/ayon_resolve/plugins/publish/collect_editorial_package.py
robin-ynput: I couldn't find `folderEntity` in neither `instance.data` nor `instance.context.data`, so leave it as it is for now.

client/ayon_resolve/plugins/create/create_editorial_package.py
```diff
@@ -1,88 +1,155 @@
...
+
+            # TODO: backward compatibility for legacy workflow instances
+            # add variant into instance data in case it is missing in publish
+            # data
+            if "variant" not in publish_data:
+                product_name = publish_data["productName"]
+                product_type = publish_data["productType"]
+                publish_data["variant"] = product_name.split(product_type)[1]
```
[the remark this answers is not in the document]
robin-ynput: I think this is safe to leave it as is for now. If a custom anatomy template was used and `editorial_pkg` was not created with the expected `{productName}{variant}` product name, then it'll still be properly detected.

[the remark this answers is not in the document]
BigRoy: Actually, that'd be an `IndexError` though?

```python
>>> "helloWorld".split("pkg")[1]
  File "<stdin>", line 1, in <module>
IndexError: list index out of range
```

Yet still - may be an edge case.

client/ayon_resolve/plugins/publish/collect_editorial_package.py
BigRoy: You are right - the folder entity gets collected here: https://github.com/ynput/ayon-core/blob/ab24518f96ecaed74593f32e10448122105b73e7/client/ayon_core/plugins/publish/collect_anatomy_instance_data.py#L40 which is at a later `order`

PR #61, opened 2025-06-04
Make ayon-resolve support free DaVinci Resolve version.

rdelillo:
## Changelog Description

Rework the Fusion+Resolve application gathering so it can accomodate the free version of DaVinci Resolve. 
Based on: https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=113252

## Testing notes:
1. Tested on DaVinci Resolve 20 and DaVinci Resolve Studio 20
2. Ensure all apps are loaded as expcted
3. Publish editorial data from free version

### DIFF
--- a/client/ayon_resolve/api/pipeline.py
+++ b/client/ayon_resolve/api/pipeline.py
@@ -32,7 +32,10 @@ from ayon_core.host import (
 
 from . import constants
 from . import lib
-from .utils import get_resolve_module
+from .utils import (
+    get_resolve_module,
+    set_resolve_module
+)
 from .workio import (
     open_file,
     save_file,
@@ -56,6 +59,7 @@ AVALON_CONTAINERS = ":AVALON_CONTAINERS"
 
 class ResolveHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
     name = "resolve"
+    _app = None
 
     def install(self):
         """Install resolve-specific functionality of avalon-core.
@@ -79,11 +83,28 @@ class ResolveHost(HostBase, IWorkfileHost, ILoadHost, IPublishHost):
         register_creator_plugin_path(CREATE_PATH)
         register_inventory_action_path(INVENTORY_PATH)
 
+        # DaVinci Resolve version >= 20
+        # Set api resolve modules from undocumented injected cached app.
+        # https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=113252
+        try:
+            bmdvf = self._app
+            bmdvr = self._app.GetResolve()
+            assert bmdvr and bmdvf
+            set_resolve_module(bmdvr, bmdvf)
+
+        # If any issue, default to DaVinci Resolve Studio mechanism.
+        except Exception as error:
+            log.info(
+                "Could not gather resolve apps from cached entry point %r. "
+                "Default to DaVinci Resolve Studio specific logic.",
+                error,
+            )
+            get_resolve_module()
+
         # register callback for switching publishable
         pyblish.register_callback("instanceToggled",
                                   on_pyblish_instance_toggled)
 
-        get_resolve_module()
 
     def open_workfile(self, filepath):
         success = open_file(filepath)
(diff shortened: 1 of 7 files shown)


client/ayon_resolve/startup.py
```diff
@@ -27,6 +27,13 @@ def ensure_installed_host():
...
         return host
 
+    # Register injected "app" variable at class level for future uses.
+    # For free version of DaVinci Resolve, this seems to be
+    # the only way to gather the Resolve/Fusion applications.
+    #
+    # https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=113252
+    ayon_resolve.api.ResolveHost.set_resolve_modules_from_app(app)  # noqa: F821
+
     host = ayon_resolve.api.ResolveHost()
```
BigRoy, line 37: Does this work?
```suggestion
    host = ayon_resolve.api.ResolveHost()
    host.set_resolve_modules_from_app(app)
```

Seems more readable. :)
Also, what you could do at the top of the file is maybe:
```python
# App is provided when running a Resolve script from the menu
app: object   # noqa: F821
```
Or something along those lines.
Just to indicate more clearly that it comes from somewhere - provided by e.g. Resolve running the script.

client/ayon_resolve/api/pipeline.py
```diff
@@ -79,11 +83,28 @@ def install(self):
         register_creator_plugin_path(CREATE_PATH)
         register_inventory_action_path(INVENTORY_PATH)
 
+        # DaVinci Resolve version >= 20
+        # Set api resolve modules from undocumented injected cached app.
+        # https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=113252
+        try:
+            bmdvf = self._app
+            bmdvr = self._app.GetResolve()
+            assert bmdvr and bmdvf
```
BigRoy, line 92: Is `assert` the right approach here? Since they can be skipped in certain python run modes, etc. etc.
I know @iLLiCiTiT always likes to slap me in the face about this so can likely explain better.

client/ayon_resolve/startup.py
rdelillo: So I've added your suggestion with and move up some documentation so this info is really straight in your face.
(ruff still complains about further `app` usage down the road, not sure why).
```
app: object   # noqa: F821
```

Regarding the code readability, I'd prefer stick with my implementation if that's OK.
I think it better enforces that this is something that needs to be cached immediatly at class level before any instance is made. Because if you don't, you might loose any possibility to create workable instance down the road. 
Your suggestion looks like we are setting an attribute instance instead.

jakubjezek001 (approved):
I am not able to test it since I do not have the Free version, but code seems valid. Anybody want to try? https://swr.cloud.blackmagicdesign.com/DaVinciResolve/v20.0/DaVinci_Resolve_20.0_Windows.zip?verify=1749114089-lewdNVU86N8GMzKeY%2F2kZfqk8uSM8clhuqBwFEsEHso%3D

MustafaJafar (approved):
This is nice! 

- I can open all the pipeline tools
image
- Workfile tool works, I Can save, change context and upversion. 
image
- Loading and scene inventory work
image
- creating instances work, triggering validation work. I was not able to publish, I guess  becasue of my limited knewledge. (log mentioned `Could not retrieve otioClip for shot %r`)
image



----
<details><summary>I got this error when loading.. (I suspect it happens becasue I don't have a user named libor on my instance, so ignore it).</summary>

```
During load error happened on Product: "renderGenericMain_sh020" Representation: "h264_png" Version: 6

Error message: 'author'

  File "E:\Ynput\ayon-core\client\ayon_core\tools\loader\models\actions.py", line 740, in _load_representations_by_loader
    load_with_repre_context(
  File "E:\Ynput\ayon-core\client\ayon_core\pipeline\load\utils.py", line 319, in load_with_repre_context
    return loader.load(repre_context, name, namespace, options)
  File "E:\Ynput\ayon-resolve\client\ayon_resolve\plugins\load\load_media.py", line 217, in load
    item = self._import_media_to_bin(context, media_pool, representation)
  File "E:\Ynput\ayon-resolve\client\ayon_resolve\plugins\load\load_media.py", line 277, in _import_media_to_bin
    data = self._get_container_data(context)
  File "E:\Ynput\ayon-resolve\client\ayon_resolve\plugins\load\load_media.py", line 386, in _get_container_data
    data[key] = version["data"][key]
KeyError: 'author'
```
I still can see the product loaded but I can't find it in the scene inventory.
image

</details>

PR #63, opened 2025-06-17
load audio product types

tweak-wtf:
## Changelog Description
This adds support for loading and managing audio products via AYON Loader.

This is rather a patch for `load_media` as it handles audio just as good as imagery.

We might think about deprecating `load_clip` as it's currently throwing errors and all of my editors are always using `load_media` over it. Thie difference is that `load_clip` places metadata on the TimelineItem whereas `load_media` places them on the MediaSource and I'm not quite sure in which scenarios the former would be desireable

## Testing notes:
1. ensure an `audio` product is published to your project
2. open Resolve
3. locate and load the `audio` product via AYON Loader
4. play around with the Scene Inventory features

### DIFF
--- a/client/ayon_resolve/plugins/load/load_media.py
+++ b/client/ayon_resolve/plugins/load/load_media.py
@@ -137,11 +137,11 @@ def find_clip_usage(media_pool_item, project=None):
 class LoadMedia(LoaderPlugin):
     """Load product as media pool item."""
 
-    product_types = {"render2d", "source", "plate", "render", "review"}
+    product_types = {"render2d", "source", "plate", "render", "review", "audio"}
 
     representations = ["*"]
     extensions = set(
-        ext.lstrip(".") for ext in IMAGE_EXTENSIONS.union(VIDEO_EXTENSIONS)
+        ext.lstrip(".") for ext in IMAGE_EXTENSIONS.union(VIDEO_EXTENSIONS, {".wav"})
     )
 
     label = "Load media"


client/ayon_resolve/plugins/load/load_media.py
```diff
@@ -137,11 +137,11 @@ def find_clip_usage(media_pool_item, project=None):
 class LoadMedia(LoaderPlugin):
     """Load product as media pool item."""
 
-    product_types = {"render2d", "source", "plate", "render", "review"}
+    product_types = {"render2d", "source", "plate", "render", "review", "audio"}
 
     representations = ["*"]
     extensions = set(
-        ext.lstrip(".") for ext in IMAGE_EXTENSIONS.union(VIDEO_EXTENSIONS)
+        ext.lstrip(".") for ext in IMAGE_EXTENSIONS.union(VIDEO_EXTENSIONS, {".wav"})
```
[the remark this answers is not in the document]
tweak-wtf: yeah i would suggest to add `AUDIO_EXTENSIONS` to `ayon_core.lib.transcoding` as a constant.
maybe this could be used by other addons... looking at premiere/flame/hiero

BigRoy (approved):
LGTM

client/ayon_resolve/plugins/load/load_media.py
```diff
@@ -24,6 +24,7 @@
 
 
 FRAME_SPLITTER = "__frame_splitter__"
+RESOLVE_AUDIO_EXTENSIONS = {".wav", ".aif", ".aiff", ".mp3", ".flac", ".aac", ".m4a", ".ac3"}
```
rdelillo, line 27: Is that all of the audio extension Resolve support or all of the audio you can think of ?
Not related to this PR, I'm just wondering if we should move it to `ayon-core` at some point:
https://github.com/ynput/ayon-core/blob/4e02667b898aeeaccfdf44de8b145df241e34caa/client/ayon_core/lib/transcoding.py#L47

tweak-wtf: i wanted to add `AUDIO_EXTENSIONS` to ayon-core, however @BigRoy pointed out that the next core release might take some time. 
so in order to get this merged faster he suggested implementing it here.

i'd be also cool with waiting for a core update. in the end i'm releasing that stuff internally right away anyways. just contributing back

rdelillo: yeah.. we can keep it there, release and log an issue in ayon-core I believe.

PR #65, opened 2025-06-20
adds product_type `image` to LoadMedia

tweak-wtf:
## Changelog Description
Adds `image` to supported product types of `LoadMedia` loader plugin. 

So far this is a oneliner. We're testing but so far it seems this was maybe just missed at some point.

## Testing notes:
1. launch Loader in Resolve
2. navigate to an `image` product type
3. rightclick, load it


BigRoy (approved):
Yes. 🚀

PR #71, opened 2025-09-23
Create: Use kwargs to call product name functions

iLLiCiTiT:
## Changelog Description
Use kwargs to call product name functions and use cached data to call them if possible.

This is preparation for future change in `get_product_name` functionality and update of already added change in `get_product_name` method.

## Testing notes:
1. Product names are calculated correctly.

### DIFF
--- a/client/ayon_resolve/plugins/create/create_workfile.py
+++ b/client/ayon_resolve/plugins/create/create_workfile.py
@@ -49,16 +49,18 @@ class CreateWorkfile(AutoCreator):
         variant = self.default_variant
         project_name = self.create_context.get_current_project_name()
         host_name = self.create_context.host_name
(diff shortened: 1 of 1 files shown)


client/ayon_resolve/plugins/create/create_workfile.py
```diff
@@ -49,16 +49,18 @@ def _create_new_instance(self):
...
         task_entity = self.create_context.get_current_task_entity()
         folder_path = folder_entity["path"]
         task_name = task_entity["name"]
         product_name = self.get_product_name(
-            project_name,
-            folder_entity,
-            task_entity,
-            self.default_variant,
-            host_name,
+            project_name=project_name,
```
jakubjezek001, line 58: Is it necessary to add project name if project_entity is added too?

PR #74, opened 2025-11-19
Gh48 resolve breaks when not installed in opt

vincentullmann:
draft PR while I'm waiting for user feedback

## Changelog Description
Add support for a custom `RESOLVE_ROOT` environment variable to specify resolve install location

This change introduces a check for a `RESOLVE_ROOT` environment variable to determine Resolves installation directory. If the variable is not defined, the logic falls back to the default path: `/opt/resolve`.

## Testing notes:
1. install resolve in a non standard location
2. define `RESOLVE_ROOT` to point to that location
3. launch resolve.

## To be discussed:
For the current implemented I've picked the simplest approach to allow user testing.
However while revieing I started wondering whether it might be more flexible to allow users to define `RESOLVE_SCRIPT_API`, `RESOLVE_SCRIPT_LIB`, and `RESOLVE_UTILITY_SCRIPTS_DIR` directly instead introducing a new variable.
```py
        if not self.launch_context.env.get("RESOLVE_SCRIPT_API"):
            # current logic 
            # ...
            #
            self.launch_context.env["RESOLVE_SCRIPT_API"] = resolve_script_api.as_posix()

        if not self.launch_context.env.get("RESOLVE_SCRIPT_LIB"):
            # current logic 
            # ...
            #
            self.launch_context.env["RESOLVE_SCRIPT_LIB"] = resolve_script_lib.as_posix()
```
This would provide more granular control while still preserving backward compatibility. Open to feedback on whether this direction makes more sense.

### DIFF
--- a/client/ayon_resolve/api/utils.py
+++ b/client/ayon_resolve/api/utils.py
@@ -40,7 +40,8 @@ def get_resolve_module():
                 )
             )
         elif sys.platform.startswith("linux"):
-            expected_path = "/opt/resolve/libs/Fusion/Modules"
+            resolve_root = os.getenv("RESOLVE_ROOT", "/opt/resolve")
+            expected_path = f"{resolve_root}/libs/Fusion/Modules"
         else:
             raise NotImplementedError(
                 "Unsupported platform: {}".format(sys.platform)


client/ayon_resolve/api/utils.py
```diff
@@ -40,7 +40,8 @@ def get_resolve_module():
                 )
             )
         elif sys.platform.startswith("linux"):
-            expected_path = "/opt/resolve/libs/Fusion/Modules"
+            resolve_root = os.getenv("RESOLVE_ROOT", "/opt/resolve")
+            expected_path = f"{resolve_root}/libs/Fusion/Modules"
```
BigRoy: I think this would need to apply to all platforms - not just linux. Right?
Or is there a reason why this _may_ only apply to Linux?

PR #83, opened 2026-03-05
🏛️Product base types: Product types and base types

iLLiCiTiT:
## Changelog Description
Use product base types and add support for product types used as aliases.

Codebase of Resolve addon is fully using product base type as "pipeline" type and product type as an alias. Load plugins define `product_base_types` for filtering, create plugins do define `product_base_type` and the whole logic is expecting that product base type is being used in the system.

Shot create plugin have option to define custom product types for plate and audio instances.

Side changes:
- Moved override of `apply_settings` to the only create plugin which actually uses it.
- Removed some backwards compatibily (classes `Creator` and `PublishClip`).
- Use 'product' instead of 'subset'.
- Marked base classes to be skipped during discovery.
- Don't use "safe getters" for settings to make sure settings are actually used.

## Testing notes:
This is untested PR from my side, please make sure it is fully reviewed.
### Basics
1. Use ayon-core >= 1.8.0 or current develop if is not released yet.
2. Upload the addon to server to use new settings model.
3. Existing workfiles should load up and should be possible to publish.
4. Creating new instances should work as expected.

### Product types
1. Define product types for plate and audio in settings.
2. Publisher should show selection of them in pre-create attributes.
3. Create them.
5. Product name should use the product type in name (if template defines that).
6. Publish the instances.
7. Loader tool should show the product type and product base type in UI.

### DIFF
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -298,7 +298,7 @@ class ResolveCreator(Creator):
 
     def apply_settings(self, project_settings):
         resolve_create_settings = (
-            project_settings.get("resolve", {}).get("create")
+            project_settings[self.settings_category]["create"]
         )
         self.presets = resolve_create_settings.get(
             self.__class__.__name__, {}
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -304,7 +304,7 @@ class ResolveCreator(Creator):
             self.__class__.__name__, {}
         )
 
-    def create(self, subset_name, instance_data, pre_create_data):
+    def create(self, product_name, instance_data, pre_create_data):
         # adding basic current context resolve objects
         self.project = lib.get_current_resolve_project()
         self.timeline = lib.get_current_timeline()
@@ -510,12 +510,12 @@ class PublishableClip:
             for key in ["folder", "episode", "sequence", "track", "shot"]
         }
 
-        # build subset name from layer name
+        # build product name from layer name
         if self.variant == "<track_name>":
             self.variant = self.track_name
 
-        # create subset for publishing
-        # TODO: Use creator `get_subset_name` to correctly define name
+        # create product name for publishing
+        # TODO: Use creator `get_product_name` to correctly define name
         self.product_name = self.product_base_type + self.variant.capitalize()
 
     def _replace_hash_to_expression(self, name, text):
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -312,10 +312,6 @@ class ResolveCreator(Creator):
             self.selected = lib.get_current_timeline_items(filter=False)
 
 
-# alias for backward compatibility
-Creator = ResolveCreator  # noqa
-
-
 class PublishableClip:
     """
     Convert a track item to publishable instance
@@ -729,9 +725,6 @@ class PublishableClip:
         for key in par_split:
             parent = self._convert_to_entity(key)
             self.parents.append(parent)
-
-# alias for backward compatibility
-PublishClip = PublishableClip  # noqa
 
 
 class HiddenResolvePublishCreator(HiddenCreator):
--- a/client/ayon_resolve/api/plugin.py
+++ b/client/ayon_resolve/api/plugin.py
@@ -294,15 +294,6 @@ class ResolveCreator(Creator):
     settings_category = "resolve"
     host_name = "resolve"
     marker_color = "Purple"
-    presets = {}
-
-    def apply_settings(self, project_settings):
-        resolve_create_settings = (
-            project_settings[self.settings_category]["create"]
-        )
-        self.presets = resolve_create_settings.get(
-            self.__class__.__name__, {}
-        )
 
     def create(self, product_name, instance_data, pre_create_data):
         # adding basic current context resolve objects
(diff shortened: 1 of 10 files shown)


client/ayon_resolve/api/plugin.py
```diff
@@ -511,13 +501,13 @@ def get(key):
...
+        # build product name from layer name
         if self.variant == "<track_name>":
             self.variant = self.track_name
 
-        # create subset for publishing
-        # TODO: Use creator `get_subset_name` to correctly define name
-        self.product_name = self.product_type + self.variant.capitalize()
+        # create product name for publishing
+        # TODO: Use creator `get_product_name` to correctly define name
+        self.product_name = self.product_base_type + self.variant.capitalize()
```
rdelillo: I had to revert this to make it work for me. Not sure if this is right but `PublishableClip.product_base_type` does not exist.
```suggestion
        self.product_name = self.product_type + self.variant.capitalize()
```

rdelillo (commented):
Tested locally, the following worked with the change I've suggested:
* [X] publish shot clip (create + publish)
* [X] editorial package (create + publish)
* [X] workfile (create + publish)
* [X] load clip
* [X] load editorial package
* [X] load media

client/ayon_resolve/api/plugin.py
iLLiCiTiT: Changed `product_type` to `plate_product_type` which is what it actually holds.

client/ayon_resolve/api/plugin.py
```diff
@@ -489,35 +479,38 @@ def _populate_attributes(self):
...
         self.hero_track = self.hero_track.replace(" ", "_")
-        self.review_source = (
-            get("reviewableSource") or self.review_source_default)
+        self.review_source = get(
+            "reviewableSource", self.review_source_default
+        )
 
         self.hierarchy_data = {
-            key: get(key) or self.timeline_item_default_data[key]
+            key: get(key, self.timeline_item_default_data[key])
```
jakubjezek001: This might not be related but this is happening to me now:


```
  File "C:\CODE\__YNPUT\ayon-core\client\ayon_core\pipeline\create\context.py", line 2344, in _create_with_unified_error
    result = creator.create(*args, **kwargs)
  File "C:\CODE\__YNPUT\ayon-resolve\client\ayon_resolve\plugins\create\create_shot_clip.py", line 617, in create
    publish_clip = PublishableClip(
  File "C:\CODE\__YNPUT\ayon-resolve\client\ayon_resolve\api\plugin.py", line 407, in __init__
    self._populate_attributes()
  File "C:\CODE\__YNPUT\ayon-resolve\client\ayon_resolve\api\plugin.py", line 502, in _populate_attributes
    self.hierarchy_data = {
  File "C:\CODE\__YNPUT\ayon-resolve\client\ayon_resolve\api\plugin.py", line 503, in <dictcomp>
    key: get(key, self.timeline_item_default_data[key])
KeyError: 'folder'
```

client/ayon_resolve/api/plugin.py
```diff
@@ -489,35 +479,38 @@ def _populate_attributes(self):
...
+        # build product name from layer name
         if self.variant == "<track_name>":
             self.variant = self.track_name
 
-        # create subset for publishing
-        # TODO: Use creator `get_subset_name` to correctly define name
-        self.product_name = self.product_type + self.variant.capitalize()
+        # create product name for publishing
+        # TODO: Use creator `get_product_name` to correctly define name
+        self.product_name = self.product_base_type + self.variant.capitalize()
```
jakubjezek001:
```suggestion
        self.product_name = self.plate_product_type + self.variant.capitalize()
```
this is inherited from https://github.com/ynput/ayon-resolve/pull/83/changes#diff-dbca93e0046aac9618a03741fb59f6e8795e66126b815774e9eaa88f1edfc3adR492

client/ayon_resolve/api/plugin.py
jakubjezek001: Actually looking into code, I am not sure how this could work before. The default value points only to `_folder_`

jakubjezek001 (approved):
I could not resist to commit some enhancements and fixes @iLLiCiTiT appologize for interfering with your great work. This was working for me and the position at Creator attributes is much better this way.

client/ayon_resolve/api/plugin.py
jakubjezek001: fixed in the commit

client/ayon_resolve/plugins/create/create_shot_clip.py
```diff
@@ -470,25 +493,56 @@ def header_label(text):
...
+                items=[base_type],
+                default=base_type,
+            )
+        return EnumDef(
+            def_key,
+            label=f"{type_label} Product Type",
+            tooltip=f"Select product type for {base_type}",
+            items=product_types,
+            default=product_types[0],
+        )
```
rdelillo, line 536: picky, but just to save a few lines

```suggestion
        return EnumDef(
            def_key,
            label=f"{type_label} Product Type",
            tooltip=f"Select product type for {base_type}",
            items=product_types if product_types else [base_type],
            default=product_types[0] if product_types else base_type,
        )
```

rdelillo (approved):
Approved. Retested with latest changes. Works for me.

iLLiCiTiT:
@rdelillo please do one quick re-test of creation.

rdelillo:
> @rdelillo please do one quick re-test of creation.

Confirmed it works ! Thanks @iLLiCiTiT

PR #91, opened 2026-03-29
Fallback to SetMetadata if SetClipProperty fails

### LINKED ISSUE #90: YN-0619: Resolve: allow "SetMetadata" properties in load metadata mapping
### Please describe the feature you have in mind and explain what the current shortcomings are?

Resolve Loaders allow to customize in settings additional metadata that can be set on load. It uses the "SetClipProperty" method on the media pool item, but it doesn't allow to set all types of metadata because some should go through "SetMetadata".

  

Can we allow the customizable properties to try going through "SetClipProperty" first - and if it fails to set it, then try setting as "SetMetadata" so we can customize with more properties like "Move", "Shot Type", "Environment". 

### Additional context:
link to discussion on [Discord](https://discord.com/channels/517362899170230292/1487087942500487169)
_(might be a private channel)_

_This issue was automatically created from AYON ticket [YN-0619]( https://support.ayon.app/projects/Active_Clients/overview?project=Active_Clients&type=folder&id=b284151029eb11f1b8cee94c937373d0)

jrsndl:
When applying metadata to media pool items in DaVinci Resolve, some properties may not be accepted by the SetClipProperty method. This adds a fallback attempt using SetMetadata to ensure more metadata fields are successfully applied.

Fix https://github.com/ynput/ayon-resolve/issues/90

## Testing notes:
1. In Resolve Metadata mapping, set some value for properties Take and Move
2. Both properties should be set when loading / updating new media to Resolve

### DIFF
--- a/client/ayon_resolve/plugins/load/load_media.py
+++ b/client/ayon_resolve/plugins/load/load_media.py
@@ -416,15 +416,20 @@ class LoadMedia(LoaderPlugin):
             clip_property = meta_item["name"]
             value = meta_item["value"]
             value_formatted = StringTemplate(value).format_strict(context)
-            success = media_pool_item.SetClipProperty(
-                clip_property, value_formatted)
-            if not bool(success):
-                try:
-                    media_pool_item.SetMetadata(clip_property, value_formatted)
-                except Exception as e:
+            is_set = media_pool_item.SetClipProperty(
+                clip_property,
+                value_formatted
+            )
+            if not is_set:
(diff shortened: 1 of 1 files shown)


BigRoy (approved):

BigRoy:
@jakubjezek001 works for me - can you confirm this change makes any sense to do like this?

jakubjezek001 (approved):
thanks @jrsndl

PR #93, opened 2026-03-30
Intermediate presets for Editorial Package

jrsndl:
## Changelog Description
Implement profile-based settings for intermediate representation exports

Allows configuring Resolve intermediate export settings, such as file format, codec, and OTIO options, via project profiles filtered by task type. This replaces hardcoded defaults with dynamic settings defined in the project configuration.

## Testing notes:
1. make sure the AYON Intermediate default still works
2. check that profile based settings work


jrsndl:
Here is a suggested Settings gui:
image

server/settings.py
```diff
@@ -74,11 +98,84 @@ class CreateShotClipModels(BaseSettingsModel):
...
+    file_format: str = SettingsField(
+        default_factory=list,
+        title="File Format",
+        enum_resolver=intermediate_format_enum
+    )
+    codec: str = SettingsField(
+        "",
+        title="Codec"
+    )
+
```
jakubjezek001, line 160:
```suggestion
    task_types: list[str] = SettingsField(
        default_factory=list,
        title="Task types",
        enum_resolver=task_types_enum
    )
    task_names: list[str] = SettingsField(
        default_factory=list,
        title="Task names"
    )
    path: str = SettingsField(
        "",
        title="Path to output preset",
        section="Resolve preset properties",
    )
    file_format: str = SettingsField(
        default_factory=list,
        title="File Format",
        enum_resolver=intermediate_format_enum
    )
    codec: str = SettingsField(
        "",
        title="Codec"
    )
    export_otio: bool = SettingsField(
        title="Export OTIO",
        description="When enabled AYON will export OTIO file along with intermediate file.",
        section="OTIO file properties",
    )
    otio_rootless: bool = SettingsField(
        title="Use rootless OTIO paths",
        description="When enabled AYON will convert all paths in OTIO to be rootless.",
    )
```

jakubjezek001:
> Here is a suggested Settings gui

Hi @jrsndl thanks for the proposed preset schema. See suggested changes. Also since the `Default preset` is sort of implicit it would be better to actually completely ditch it and add its content into default preset, so anybody can see what is it about and perhaps tune it to their needs.

jrsndl:
> > Here is a suggested Settings gui
> 
> Hi @jrsndl thanks for the proposed preset schema. See suggested changes. Also since the `Default preset` is sort of implicit it would be better to actually completely ditch it and add its content into default preset, so anybody can see what is it about and perhaps tune it to their needs.

Having editable default preset would be great. But it requires a path to the Resolve settings xml distributed by AYON. Maybe in a similar way how you now distribute default OCIO? I am not sure how to do this properly...

jakubjezek001:
> Maybe in a similar way how you now distribute default OCIO? I am not sure how to do this properly...

Yes, something like this could be good approach. I would recommend this solution:

## Default preset
Name: `AYON_intermediates`
Path to output preset: `{RESOLVE_ADDON_ROOT}/presets/render/AYON_intermediates.xml`

It would of course require to add `RESOLVE_ADDON_ROOT` key into formatting data and make sure the `preset_path` is always formatted with context anatomy keys (some presets might be part of `/ProjectName/resources/resolve/presets` path so they might need to use `{root[work]}/{project[name]}/resources/resolve/presets/foo.xml`). For the formatting I would suggest to use ayon-core methods:

```python
from copy import deepcopy
from ayon_core.lib import StringTemplate

from ayon_resolve.utils import RESOLVE_ADDON_ROOT

formatting_data = deepcopy(anatomy_data)
formatting_data.update({
     "RESOLVE_ADDON_ROOT":  RESOLVE_ADDON_ROOT
})

preset_path_formatted = StringTemplate.format_strict_template(
    preset_path, formatting_data
)
```

server/settings.py
```diff
@@ -84,11 +108,107 @@ class CreateShotClipModels(BaseSettingsModel):
...
+    )
+    file_format: str = SettingsField(
+        default_factory=list,
+        title="File Format",
+        enum_resolver=intermediate_format_enum
+    )
+    codec: str = SettingsField(
+        "",
+        title="Codec"
+    )
```
jakubjezek001, line 149: it seems this is duplicity

server/settings.py
```diff
@@ -84,11 +108,107 @@ class CreateShotClipModels(BaseSettingsModel):
...
+    task_types: list[str] = SettingsField(
+        default_factory=list,
+        title="Task types",
+        enum_resolver=task_types_enum
+    )
+    task_names: list[str] = SettingsField(
+        default_factory=list,
+        title="Task names"
+    )
+    path: str = SettingsField(
```
jakubjezek001: In the plugin code this is called `preset_path`

PR #104, opened 2026-06-12
Allow a profile to create multiple render passes

tweak-wtf:
## Changelog Description
This draft shall make it possible to render more than 1 representation per instance.
Also it aims at making source clip integration optional.

General goal of this is to make use of resolve rendering engine as much as possible to not needing to go through `extract_review` as this comes with performance penalties compared to resolve.

- changes profile structure to allow multiple representations to be rendered
- adds `Upload as reviewable` tag (really just `webreview` under the hood)

I tested this for `plate` and `editorial_pkg` products on windows only.

## Testing notes:
- build & upload the addon
- configure more than 1 representation pass
- publish and check for your expected output

### DIFF
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -1,4 +1,5 @@
 import os
+from typing import Any
 from pathlib import Path
 from pprint import pformat
 
@@ -48,30 +49,37 @@ class ExtractProductResources(
 
     def process(self, instance):
         instance.data.setdefault("representations", [])
+        integrate_clip_source, all_output_settings = self.get_settings(instance)
 
-        settings = self.get_settings(instance)
-        preset_path = Path(self.resolve_preset_path(settings["preset_path"]))
-        product_base_type = instance.data["productBaseType"]
+        # todo: find the source of this, i think it's some otio extractor
+        #       maybe skip adding the repre there? but settings would be in a different extractor... doesn't make sense
+        if not integrate_clip_source:
+            instance.data["representations"] = []   # disables clip source integration
 
-        # set rendering logger to inherit from publisher's logger
-        rendering.log = self.log
+        for settings in all_output_settings:
+            preset_path = Path(self.resolve_preset_path(settings["preset_path"]))
+            product_base_type = instance.data["productBaseType"]
 
-        if product_base_type == "editorial_pkg":
-            self._process_editorial_pkg(instance, settings, preset_path)
-        elif product_base_type == "plate":
-            self._process_plate(instance, settings, preset_path)
-        else:
-            self.log.warning(
-                "ExtractProductResources: unhandled product base type '%s', skipping.", product_base_type
-            )
+            # set rendering logger to inherit from publisher's logger
+            rendering.log = self.log
 
-    def get_settings(self, instance):
+            if product_base_type == "editorial_pkg":
+                self._process_editorial_pkg(instance, settings, preset_path)
+            elif product_base_type == "plate":
+                self._process_plate(instance, settings, preset_path)
+            else:
+                self.log.warning(
+                    "ExtractProductResources: unhandled product base type '%s', skipping.", product_base_type
+                )
+                return
+
+    def get_settings(self, instance) -> list[dict[str, Any]]:
         """Return normalised render settings for *instance*.
 
         Looks up ``resolve → publish → ExtractProductResources → profiles``,
         then profile-matches on task type / task name / product_base_type.
 
-        Returns a flat dict with keys:
+        Returns a list of flat dicts with keys:
             ``file_format``, ``codec``, ``preset_path``,
             and for *editorial_pkg* only: ``export_otio``, ``otio_rootless``.
         """
@@ -101,26 +109,31 @@ class ExtractProductResources(
             return self.get_default_settings(product_base_type)
 
         self.log.debug(f"Matched preset: {profile.get('name')}")
-        return self._normalize_preset(profile, product_base_type)
 
-    def _normalize_preset(self, preset, product_base_type):
+        result = []
+        for preset in profile.get(product_base_type, []):
+            normalized = self._normalize_preset(preset)
+            result.append(normalized)
+
+        return (profile["integrate_clip_source"], result)
+
+    def _normalize_preset(self, preset):
         """Flatten the nested preset structure into a render-ready dict."""
-        sub = preset.get(product_base_type, {})
-        preset_type = sub.get("preset_type")
-        fmt = sub.get(preset_type, {})
+        preset_type = preset["settings"].get("preset_type")
+        fmt = preset["settings"].get(preset_type, {})
 
         normalized = {
-            "name":        preset["name"],
+            "name":        preset["shared"]["repre_name"],
             "file_format": fmt.get("format"),
             "codec":       fmt.get("codec"),
             "preset_path": fmt.get("preset_path"),
-            "with_handles": sub.get("with_handles"),
-            "tags":        preset["tags"],
-            "custom_tags": preset["custom_tags"],
+            "with_handles": preset["settings"].get("with_handles"),
+            "tags":        preset["shared"]["tags"],
+            "custom_tags": preset["shared"]["custom_tags"],
+            "export_otio": preset["settings"].get("export_otio", None),
+            "otio_rootless": preset["settings"].get("otio_rootless", None),
         }
-        if product_base_type == "editorial_pkg":
-            normalized["export_otio"] = sub.get("export_otio")
-            normalized["otio_rootless"] = sub.get("otio_rootless")
+        self.log.debug(f"{normalized = }")
         return normalized
 
     def get_default_settings(self, product_base_type="editorial_pkg"):
@@ -302,17 +315,16 @@ class ExtractProductResources(
 
         frame_start = instance.data["frameStart"]
 
-        with_handles = settings.get("with_handles", False)
-
         folder_slug = instance.data["folderPath"].lstrip("/").replace("/", "_")
         clip_name = timeline_item.GetName()
         staging_dir = (
-            Path(self.staging_dir(instance)) / f"{folder_slug}_{clip_name}"
+            Path(self.staging_dir(instance)) / f"{folder_slug}_{clip_name}_{settings['name']}"
         )
         staging_dir.mkdir(parents=True, exist_ok=True)
         self.log.info("Staging directory: %s", staging_dir)
 
         preset_data = {}
+        with_handles = settings.get("with_handles", False)
         if with_handles:
             preset_data.update({
                 "NumFramesOfHandles": max(available_head, available_tail)
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -84,11 +84,6 @@ class ExtractProductResources(
             and for *editorial_pkg* only: ``export_otio``, ``otio_rootless``.
         """
         entity = get_current_task_entity()
-        if not entity:
-            self.log.warning("No current task entity — using default settings.")
-            product_base_type = instance.data["productBaseType"]
-            return self.get_default_settings(product_base_type)
-
         task_type = entity.get("taskType")
         task_name = entity.get("taskName")
         product_base_type = instance.data["productBaseType"]
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -98,10 +98,12 @@ class ExtractProductResources(
         )
 
         if not profile:
-            self.log.debug(
-                "No preset matched for family='%s', task_type='%s', task_name='%s'. Using defaults.", product_base_type, task_type, task_name
+            raise RuntimeError(
+                (
+                    f"No matching preset found for family='{product_base_type}'. "
+                    "Please visit your server settings and add a matching preset for this family."
+                )
             )
-            return self.get_default_settings(product_base_type)
 
         self.log.debug(f"Matched preset: {profile.get('name')}")
 
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -116,8 +116,8 @@ class ExtractProductResources(
 
     def _normalize_preset(self, preset):
         """Flatten the nested preset structure into a render-ready dict."""
-        preset_type = preset["settings"].get("preset_type")
-        fmt = preset["settings"].get(preset_type, {})
+        preset_type = preset["settings"]["preset_type"]
+        fmt = preset["settings"][preset_type]
 
         normalized = {
             "name":        preset["shared"]["repre_name"],
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -127,8 +127,8 @@ class ExtractProductResources(
             "with_handles": preset["settings"].get("with_handles"),
             "tags":        preset["shared"]["tags"],
             "custom_tags": preset["shared"]["custom_tags"],
-            "export_otio": preset["settings"].get("export_otio", None),
-            "otio_rootless": preset["settings"].get("otio_rootless", None),
+            "export_otio": preset["settings"].get("export_otio"),
+            "otio_rootless": preset["settings"].get("otio_rootless"),
         }
         self.log.debug(f"{normalized = }")
         return normalized
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -119,19 +119,19 @@ class ExtractProductResources(
 
     def _normalize_preset(self, preset):
         """Flatten the nested preset structure into a render-ready dict."""
-        preset_type = preset["settings"]["preset_type"]
-        fmt = preset["settings"][preset_type]
+        preset_type = preset["output_defs"]["preset_type"]
+        fmt = preset["output_defs"][preset_type]
 
         normalized = {
             "name":        preset["shared"]["repre_name"],
             "file_format": fmt.get("format"),
             "codec":       fmt.get("codec"),
             "preset_path": fmt.get("preset_path"),
-            "with_handles": preset["settings"].get("with_handles"),
+            "with_handles": preset["output_defs"].get("with_handles"),
             "tags":        preset["shared"]["tags"],
             "custom_tags": preset["shared"]["custom_tags"],
-            "export_otio": preset["settings"].get("export_otio"),
-            "otio_rootless": preset["settings"].get("otio_rootless"),
+            "export_otio": preset["output_defs"].get("export_otio"),
+            "otio_rootless": preset["output_defs"].get("otio_rootless"),
         }
         self.log.debug(f"{normalized = }")
         return normalized
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -132,6 +132,7 @@ class ExtractProductResources(
             "custom_tags": preset["shared"]["custom_tags"],
             "export_otio": preset["output_defs"].get("export_otio"),
             "otio_rootless": preset["output_defs"].get("otio_rootless"),
+            "colorspace": preset["shared"].get("colorspace"),
         }
         self.log.debug(f"{normalized = }")
         return normalized
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -133,26 +133,6 @@ class ExtractProductResources(
         self.log.debug(f"{normalized = }")
         return normalized
 
-    def get_default_settings(self, product_base_type="editorial_pkg"):
-        """Return hard-coded defaults when no matching preset is found."""
-        if product_base_type == "plate":
-            return {
-                "file_format": "EXR",
-                "codec":       "RGB half (DWAA)",
-                "preset_path": (
-                    "{ayon_render_presets}/clip/EXR_RGB_half_(DWAA).xml"
-                ),
-            }
-        return {
-            "file_format":   "QuickTime",
-            "codec":         "H.264",
-            "preset_path":   (
-                "{ayon_render_presets}/timeline/QuickTime_H264.xml"
-            ),
-            "export_otio": True,
-            "otio_rootless": True,
-        }
-
     def resolve_preset_path(self, preset_path):
         """Resolve the path to a render preset file.
 
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -56,10 +56,8 @@ class ExtractProductResources(
         instance.data.setdefault("representations", [])
         integrate_clip_source, all_output_settings = self.get_settings(instance)
 
-        # todo: find the source of this, i think it's some otio extractor
-        #       maybe skip adding the repre there? but settings would be in a different extractor... doesn't make sense
         if not integrate_clip_source:
-            instance.data["representations"] = []   # disables clip source integration
+            instance.data["representations"] = []
 
         for settings in all_output_settings:
             preset_path = Path(self.resolve_preset_path(settings["preset_path"]))
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -301,8 +301,7 @@ class ExtractProductResources(
         self.log.info("Staging directory: %s", staging_dir)
 
         preset_data = {}
-        with_handles = settings.get("with_handles", False)
-        if with_handles:
+        if settings["with_handles"]:
             preset_data.update({
                 "NumFramesOfHandles": max(available_head, available_tail)
             })
--- a/client/ayon_resolve/plugins/publish/extract_product_resources.py
+++ b/client/ayon_resolve/plugins/publish/extract_product_resources.py
@@ -5,7 +5,12 @@ from pprint import pformat
 
 import pyblish.api
 from ayon_core.lib import StringTemplate, filter_profiles
-from ayon_core.pipeline import Anatomy, get_current_project_name, publish
+from ayon_core.pipeline import (
+    Anatomy,
+    get_current_project_name,
+    publish,
+    PublishError,
+)
 from ayon_core.pipeline.context_tools import get_current_task_entity
 from ayon_resolve.api import rendering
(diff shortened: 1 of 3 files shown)


client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -101,26 +109,31 @@ def get_settings(self, instance):
...
 
-    def _normalize_preset(self, preset, product_base_type):
+        result = []
+        for preset in profile.get(product_base_type, []):
+            normalized = self._normalize_preset(preset)
+            result.append(normalized)
+
+        return (profile["integrate_clip_source"], result)
+
+    def _normalize_preset(self, preset):
```
BigRoy, line 117: Would it make more sense for this to return a dataclass? So that the `process` methods also take these dataclasses? Would probably make the code way more sane?

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -48,30 +49,37 @@ class ExtractProductResources(
...
         instance.data.setdefault("representations", [])
+        integrate_clip_source, all_output_settings = self.get_settings(instance)
 
-        settings = self.get_settings(instance)
-        preset_path = Path(self.resolve_preset_path(settings["preset_path"]))
-        product_base_type = instance.data["productBaseType"]
+        # todo: find the source of this, i think it's some otio extractor
+        #       maybe skip adding the repre there? but settings would be in a different extractor... doesn't make sense
+        if not integrate_clip_source:
+            instance.data["representations"] = []   # disables clip source integration
```
BigRoy, line 57: Isn't this just a matter of tagging the repres with a `delete` tag? 🤔

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -48,30 +49,37 @@ class ExtractProductResources(
...
-        if product_base_type == "editorial_pkg":
-            self._process_editorial_pkg(instance, settings, preset_path)
-        elif product_base_type == "plate":
-            self._process_plate(instance, settings, preset_path)
-        else:
-            self.log.warning(
-                "ExtractProductResources: unhandled product base type '%s', skipping.", product_base_type
-            )
+            # set rendering logger to inherit from publisher's logger
+            rendering.log = self.log
```
BigRoy, line 64: I know this isn't new - but this changes the module-level log variable and will keep it changed after this has run. If easy to fix, would be good to clean that up.

tweak-wtf: just tested without the module-level log override.
it causes the logs from `ayon_resolve.api` functions to not be visible in the publisher's details pane. however, the resolve console still prints them.

imho we can remove it completely. afaik other addons' api functions also don't propagate to the publisher.

does this open up the general logging can of worms?

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: tried it just now but unfortunately no :/
the main issue here is this
- i have a r3d file as source that i'm trying to publish as exr
- [Collect OTIO Subset Resources](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/collect_otio_subset_resources.py) picks up the r3d and adds it as representation
- [Extract OTIO trim longer video](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/extract_otio_trimming_video.py) kicks in failing with ffprobe error `Invalid data found when processing input`
- since the r3d is still a representation it will be integrated
  - doesn't really make sense as u don't want to integrate the cam source, only the plates thatw ere rendered from it before

since i haven't found any settings in ayon-core for those 2 otio related plugins i made this admittedly ugly af workaround.
should i add server settings for the otio plugins in ayon-core to e.g. remove `resolve` from the hosts?
if the collector never runs it will also never add the cam source as representation right 🤷

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: tbh i only used what was already there trying to be minimally invasive.

but sure. refactor everything preset related into 1 dataclass. can do that ✌️

BigRoy: Happy to do so in separate PR though @jakubjezek001 ?

tweak-wtf: alrighty, then i'll continue to adress the notes where possible and eventually set this to be ready for review.
on that note thanks for checking this out already

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -101,26 +109,31 @@ def get_settings(self, instance):
...
             "preset_path": fmt.get("preset_path"),
-            "with_handles": sub.get("with_handles"),
-            "tags":        preset["tags"],
-            "custom_tags": preset["custom_tags"],
+            "with_handles": preset["settings"].get("with_handles"),
+            "tags":        preset["shared"]["tags"],
+            "custom_tags": preset["shared"]["custom_tags"],
+            "export_otio": preset["settings"].get("export_otio", None),
+            "otio_rootless": preset["settings"].get("otio_rootless", None),
         }
```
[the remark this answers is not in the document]
tweak-wtf: hm i wouldn't add colorspace here.
i think handling colorspaces and their conversion here ends up being convoluted and not flexible.

i'm running an approach where i give artists a toggle to enable/disable "Post ColorGroup" in resolve directly before render triggers.
that way the artist can use vanilla resolve to do linearization.
but that's not in this PR. can open it up if there is interest tho

client/ayon_resolve/plugins/publish/extract_product_resources.py
BigRoy: > should i add server settings for the otio plugins in ayon-core to e.g. remove `resolve` from the hosts? if the > collector never runs it will also never add the cam source as representation right 🤷

You may want to check out https://github.com/ynput/ayon-resolve/pull/109

I think it overlaps?

client/ayon_resolve/plugins/publish/extract_product_resources.py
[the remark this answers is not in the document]
tweak-wtf: having said that we might completely remove `colorspace` from server settings.
not sure what the original intent behind this is

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -48,30 +49,37 @@ class ExtractProductResources(
...
+                    "ExtractProductResources: unhandled product base type '%s', skipping.", product_base_type
+                )
+                return
 
-    def get_settings(self, instance):
+    def get_settings(self, instance) -> list[dict[str, Any]]:
         """Return normalised render settings for *instance*.
 
         Looks up ``resolve → publish → ExtractProductResources → profiles``,
         then profile-matches on task type / task name / product_base_type.
```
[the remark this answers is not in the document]
tweak-wtf: the scenario in which task isn't available is impossible.
in order to launch resolve (or really any dcc) the artist must launch it from a task context. 

therefore i removed the ealy return if no task is present 72ff09d5d7bcf70e38645e3227600e8da97be46e
also i'm raising if no profile is found instead of returning defaults 90c9ffa0bf651f39e99e4bbb53d16812549114be.

the defaults should be in the server settings so will be adressing this there

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -101,26 +109,31 @@ def get_settings(self, instance):
             return self.get_default_settings(product_base_type)
```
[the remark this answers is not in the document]
tweak-wtf: this is resolved by removing the need to invoke `get_default_settings`.
this should be the defaults of the server settings

72ff09d5d7bcf70e38645e3227600e8da97be46e
90c9ffa0bf651f39e99e4bbb53d16812549114be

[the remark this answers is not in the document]
tweak-wtf: and the server defaults got added with 307b31b72d10327d7fdbde064d9f6c9ab0a64f0b

client/ayon_resolve/plugins/publish/extract_product_resources.py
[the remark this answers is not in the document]
tweak-wtf: default got added with 307b31b72d10327d7fdbde064d9f6c9ab0a64f0b

server/settings.py
```diff
@@ -536,33 +581,34 @@ class ResolveSettings(BaseSettingsModel):
...
-                    "custom_tags": []
-                }
-            ]
-        }
-    }
+    # todo: paste sane defaults, commented out rn to not make frontend suffer too much
+    # "publish": {
+    #     "ExtractProductResources": {
+    #         "profiles": [
+    #             {
```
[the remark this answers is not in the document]
tweak-wtf: adressed in 307b31b72d10327d7fdbde064d9f6c9ab0a64f0b

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: i tried with the changes in https://github.com/ynput/ayon-resolve/pull/109 but to no avail :/

the main issue is that [Collect OTIO Subset Resources](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/collect_otio_subset_resources.py) will always run and will always add the source media to this instance's representations.

i kinda see 3 options here:
1. make [Collect OTIO Subset Resources](https://github.com/ynput/ayon-core/blob/develop/client/ayon_core/plugins/publish/collect_otio_subset_resources.py) optional in ayon-core
2. add e.g. `skip_otio` bool to instance.data during creation and respect this in the otio subset collector
3. keep it as is in this PR

kinda leaning towards option 1 tbh but would require a new PR.

thoughts on this @BigRoy ?

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -101,26 +109,31 @@ def get_settings(self, instance):
...
 
         self.log.debug(f"Matched preset: {profile.get('name')}")
-        return self._normalize_preset(profile, product_base_type)
 
-    def _normalize_preset(self, preset, product_base_type):
+        result = []
+        for preset in profile.get(product_base_type, []):
+            normalized = self._normalize_preset(preset)
+            result.append(normalized)
```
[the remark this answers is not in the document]
tweak-wtf: adressed in e988bfb4eb02d200e1965cbf558db1a00f19ef3c

settings now validate for unique representation names per repre of each profile

client/ayon_resolve/plugins/publish/extract_product_resources.py
jakubjezek001: It seems to me that better approach here would be 3. since it is kept and controled at the profile level. The   Collect OTIO Subset Resources should still do what it does (collect representation with resources found in mediasource at OTIO clip), but this plugin should be able to override what ever it found with new representations.

client/ayon_resolve/plugins/publish/extract_product_resources.py
jakubjezek001: > does this open up the general logging can of worms?

yes it sort of does. @iLLiCiTiT any opinion about this?

iLLiCiTiT: I don't think it should be handled like this. If it should capture all logs and it does not happen then create an issue in ayon-core to fix it.

If you still insist on keeping this then I'd recommend to create context manager which will replace the log and then set it back to origin log at the end.

Another option is to be able to pass in logger object to the functions.

server/settings.py
```diff
@@ -300,33 +354,36 @@ class ProductResourcesPresetModel(BaseSettingsModel):
...
 
+    @validator("editorial_pkg", "plate")
+    def validate_unique_outputs(cls, values):
+        # ensure_unique_names unfortunately is hardcoded to use `name` as field key
+        names = []
+        for val in values:
+            if val.shared.repre_name not in names:
+                names.append(val.shared.repre_name)
+            else:
+                raise ValueError(f"Duplicate representation name: {val.shared.repre_name}")
```
iLLiCiTiT:
```suggestion
            if val.shared.repre_name in names:
                raise ValueError(
                    f"Duplicate representation name: {val.shared.repre_name}"
                )
            names.append(val.shared.repre_name)
```

server/settings.py
```diff
@@ -95,6 +95,7 @@ def representation_tags_enum():
         {"value": "review", "label": "Extract review processing"},
         {"value": "delete", "label": "Delete - as intermediate"},
         {"value": "passing", "label": "Skip Extract Review"},
+        {"value": "webreview", "label": "Upload as reviewable"},
```
iLLiCiTiT, line 102: Does it mean that this can skip ExtractReview with this change?

This will mark the representation to be used for AYON web review. Which means that all output representations from ExtractReview will have it too (even if the the output definition in the settings won't have it).

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -302,17 +292,16 @@ def _process_plate(self, instance, settings, preset_path):
...
         clip_name = timeline_item.GetName()
         staging_dir = (
-            Path(self.staging_dir(instance)) / f"{folder_slug}_{clip_name}"
+            Path(self.staging_dir(instance)) / f"{folder_slug}_{clip_name}_{settings['name']}"
         )
         staging_dir.mkdir(parents=True, exist_ok=True)
         self.log.info("Staging directory: %s", staging_dir)
 
         preset_data = {}
+        with_handles = settings.get("with_handles", False)
```
iLLiCiTiT:
```suggestion
        with_handles = settings["with_handles"]
```

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -95,54 +98,41 @@ def get_settings(self, instance):
...
+            result.append(normalized)
+
+        return (profile["integrate_clip_source"], result)
+
+    def _normalize_preset(self, preset):
         """Flatten the nested preset structure into a render-ready dict."""
-        sub = preset.get(product_base_type, {})
-        preset_type = sub.get("preset_type")
-        fmt = sub.get(preset_type, {})
+        preset_type = preset["settings"].get("preset_type")
```
iLLiCiTiT: Please don't use `.get(...)` for keys that should be always there. If I read the settings models correctly this should be just
```python
preset_type = preset["settings"]["preset_type"]
fmt = preset["settings"][preset_type]
```

server/settings.py
```diff
@@ -278,11 +280,63 @@ class PlateFormatModel(BaseSettingsModel):
...
+    )
+    shared: ProductTypeSharedModel = SettingsField(
+        default_factory=ProductTypeSharedModel,
+        title="Shared Attributes"
+    )
+
+
+class PlateModel(BaseSettingsModel):
+    _layout = "expanded"
+    settings: PlateFormatModel = SettingsField(
```
iLLiCiTiT: `settings` key (and variable) is very confusing (at least for me), raising a question if there is a better name than `settings`. Like `output_defs` (I'm not sure if it is definition of outputs?).

It is not important just for readability, also for explaining. Currently there is just `Plate` title, which really doesn't tell much to person changing the settings.

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -95,54 +98,41 @@ def get_settings(self, instance):
...
             "codec":       fmt.get("codec"),
             "preset_path": fmt.get("preset_path"),
-            "with_handles": sub.get("with_handles"),
-            "tags":        preset["tags"],
-            "custom_tags": preset["custom_tags"],
+            "with_handles": preset["settings"].get("with_handles"),
+            "tags":        preset["shared"]["tags"],
+            "custom_tags": preset["shared"]["custom_tags"],
+            "export_otio": preset["settings"].get("export_otio", None),
+            "otio_rootless": preset["settings"].get("otio_rootless", None),
```
iLLiCiTiT: I guess the `export_otio` and `otio_rootless` are available only sometimes?

But the `None` default is not needed as it is default "by default".
```suggestion
            "export_otio": preset["settings"].get("export_otio"),
            "otio_rootless": preset["settings"].get("otio_rootless"),
```

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -95,54 +98,41 @@ def get_settings(self, instance):
         )
 
         if not profile:
-            self.log.debug(
-                "No preset matched for family='%s', task_type='%s', task_name='%s'. Using defaults.", product_base_type, task_type, task_name
+            raise RuntimeError(
```
iLLiCiTiT:
```suggestion
            raise PublishError(
```

iLLiCiTiT: Does it really have to raise an error?

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -95,54 +98,41 @@ def get_settings(self, instance):
...
+                (
+                    f"No matching preset found for family='{product_base_type}'. "
+                    "Please visit your server settings and add a matching preset for this family."
+                )
```
iLLiCiTiT:
```suggestion
                "No matching preset found for product base type"
                f" '{product_base_type}'. A matching preset for this product"
                " base type has to be added in server settings."
```

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -48,39 +49,41 @@ class ExtractProductResources(
...
 
         Looks up ``resolve → publish → ExtractProductResources → profiles``,
         then profile-matches on task type / task name / product_base_type.
 
-        Returns a flat dict with keys:
+        Returns a list of flat dicts with keys:
             ``file_format``, ``codec``, ``preset_path``,
             and for *editorial_pkg* only: ``export_otio``, ``otio_rootless``.
         """
         entity = get_current_task_entity()
```
iLLiCiTiT, line 89: Why this uses `get_current_task_entity` instead of `context.data["taskEntity"]`?

BTW Why was the validation removed? Is it validated ahead in validations?

server/settings.py
tweak-wtf: adressed in d25687a920351dc518d6591e3af898879300e84b

server/settings.py
tweak-wtf: hm not really, this change just lets me add a tag to a representation defined in resolve's export preset.

I'm using this PR without anything configured for resolve in `Extract Review`. I'm completely bypassing it which is the goal of this PR.
> General goal of this is to make use of resolve rendering engine as much as possible to not needing to go through `extract_review` as this comes with performance penalties compared to resolve.

server/settings.py
tweak-wtf: yeah ur right!
i'm trying to come up with better naming

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: yeah these are only present in `Editorial Package` but not in `Plate`
adressed in 2a932513a65d9df034de2a1b0b9a93fd2efe9ca9

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: okay cool then i won't take action here :v:

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: adressed in 5b4567dc1eb1e6dc5a848b756554f96de882db51

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: adressed in af133690f30ff1068e2fd4b90837bd24be5e9f23

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: adressed in 04e4442d61b279609fd601a4a2c8ab05ab049c99

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: when this extractor runs i assume a user has either a `plate` or an `editorial_pkg` to export. if that user doesn't have any settings for them it should fail and stop the publish. so i believe yes it should raise an error.

but happy to make this a publish error :)

adressed in 75cde9206704d64928842307e3855a6939a8d1b6

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: > Why this uses get_current_task_entity instead of context.data["taskEntity"]?

can't really tell, this is part of the current codebase and was introduced here: https://github.com/ynput/ayon-resolve/commit/8e958262d4cf65ec8c5057edd9d47ab5bdc57e4d#diff-1967197894dd67b098a925d8f9b4cf1f525872550105a43ce555004a25f0e335R71

but I also don't see why this can't just be retrieved from context. @jakubjezek001 any gotchas here?

> BTW Why was the validation removed? Is it validated ahead in validations?

i removed it in 72ff09d5d7bcf70e38645e3227600e8da97be46e as i found this scenario to be impossible to reach.
AYON requires a task to open a DCC, so if i can't open resolve from a non-task context how can there ever be no task  entity collected?
but i can just revert this commit if u want. just wanted to cut down on a few lines

server/settings.py
tweak-wtf: i really liked `output_defs` so i ran with it.
also worked a bit on usability of the server settings in general. should be a bit easier to configure now.

mainly adressed in 30630d9c656a70d23d211762879fe0d5ecf39936 and bc87f481786830a02856e117c6fc85caf7770106

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: tbh i would tackle this in a different PR as it's pretty unrelated to the goal of this PR.

jakubjezek001 (approved):
Tested and with some improvements form https://github.com/bEpic-studio/ayon-resolve/pull/1 this could be merged

client/ayon_resolve/plugins/publish/extract_product_resources.py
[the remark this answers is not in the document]
tweak-wtf: adressed in 785370c2f3c62287b3650240f7fd5cbb1efa48ca
thanks jeza :)

jakubjezek001:
@iLLiCiTiT or @BigRoy can this be merged and released?

client/ayon_resolve/plugins/publish/extract_product_resources.py
```diff
@@ -48,39 +54,41 @@ class ExtractProductResources(
 
     def process(self, instance):
         instance.data.setdefault("representations", [])
+        integrate_clip_source, all_output_settings = self.get_settings(instance)
 
-        settings = self.get_settings(instance)
-        preset_path = Path(self.resolve_preset_path(settings["preset_path"]))
-        product_base_type = instance.data["productBaseType"]
+        # todo: find the source of this, i think it's some otio extractor
+        #       maybe skip adding the repre there? but settings would be in a different extractor... doesn't make sense
```
BigRoy: Can we clean up this comment, or "find this source" @jakubjezek001 ?

client/ayon_resolve/plugins/publish/extract_product_resources.py
BigRoy: @jakubjezek001 can you create issue to track this - then we can resolve this for the sake of this PR.

client/ayon_resolve/plugins/publish/extract_product_resources.py
BigRoy: @jakubjezek001 can you comment here and resolve if you're ok? Using the context `taskEntity` might be better?

BigRoy (commented):
Mostly good to solve the pending comments, but if it **works** and you're happy @jakubjezek001 then I'm ok with it. Feel free to poke Robin if you really need a more in-depth review/testing.

client/ayon_resolve/plugins/publish/extract_product_resources.py
tweak-wtf: i explained why i did it like this in this thread here:
https://github.com/ynput/ayon-resolve/pull/104#discussion_r3631042465

from that we agreed to keep this as is, so happy to just remove the comment. i know where it's coming from

tweak-wtf: adressed in 2f467567f48730c532e63bfd35cf39c2648bc484

jakubjezek001:
@rdelillo will you have a time to test and review the code?

rdelillo:
Sure. Will test this tomorrow.

client/ayon_resolve/plugins/publish/extract_product_resources.py
rdelillo: Here we go: https://github.com/ynput/ayon-resolve/issues/110

rdelillo (approved):
Overhall looks good, I just added a few fixes consolidation here: [9ac832cf849a64c65dec57479b94fa3080d402b3](https://github.com/ynput/ayon-resolve/commit/9ac832cf849a64c65dec57479b94fa3080d402b3)

* [x] Fix defaults settings `settings` -> `output_defs` in default dicts
* [x] Handle settings conversion from previous `ayon-resolve` version
* [x] Default representation name to extension
* [x] Raise when no output_def is defined in settings  

@tweak-wtf feel free to merge my commit with your branch, or we can merge mine and close this one.
@jakubjezek001 would be great if you could quickly give it another try before we merge.

tweak-wtf:
hey @rdelillo thanks for the additions :)
just merged it in

client/ayon_resolve/plugins/publish/extract_product_resources.py
iLLiCiTiT: It would have to be raw class, dataclasses are failing to be importable in publish plugins.

tweak-wtf: noted. i'll pick this up in a separate PR then or shall i adress it in here?

rdelillo: Id agree with initial comment from Roy separate PR is fine for this one.

rdelillo:
@tweak-wtf I believe this good to merge.
I'd have done it myself from this PR but looks like I don't have the permissio to push on your branch.

tweak-wtf:
> @tweak-wtf I believe this good to merge. I'd have done it myself from this PR but looks like I don't have the permissio to push on your branch.

sweet... just pulled in develop ✌️

PR #108, opened 2026-07-21
Pre launch hooks: Handle workfile path in pre-launch hook

iLLiCiTiT:
## Changelog Description
Care about `workfile_path` set in launch hooks data.

Resolve does not respect explicitly set workfile path under `"workfile_path"` on hook data which means it will try to launch the last one on open.

NOTE I did have the changes localy without a PR so I'm not sure if there was a reason why not to do this?

## Testing notes:
1. Explicitly selected workfile in launcher tool should open the selected workfile instead of last workfile.


BigRoy (approved):
Works for me.

PR #111, opened 2026-09-08
Add Resolve optional pre-hook to fails if non-studio version >= 21

rdelillo:
## Changelog Description

Resolve 21.1 removed the python integration with non-studio version, making the `ayon-resolve` addon incompatible:
https://www.reddit.com/r/davinciresolve/comments/1wafb08/davinci_resolve_211_release_notes/

```markdown
# Notes:
We have moved the ability to script in Python to the Studio version.
The Python API was being used to hack studio features into the free version.

DaVinci Resolve relies on studio license sales to pay for the engineering team.
Unlike subscriptions, which have become normal but lock people's work up unless you pay monthly, we want to offer a differentiation from charging monthly license fees.
```

Changes:
* [x] Detect current DaVinci Resolve version and flavor from executable, raise if not supported.

image

## Testing notes:
1. Ensure it still works with Resolve Studio 21.1
2. Ensure it does not work anymore with Resolve 21.1

### DIFF
--- a/client/ayon_resolve/hooks/pre_resolve_startup.py
+++ b/client/ayon_resolve/hooks/pre_resolve_startup.py
@@ -31,7 +31,7 @@ class PreLaunchResolveStartup(PreLaunchHook):
             if not is_studio and major_version >= 21:
                 raise RuntimeError(
                     f"AYON is not supported in {resolve_version} "
-                    "Only the studio version supports python external scripting: \n"
+                    "Only the Studio version supports Python external scripting:\n"
                     "https://www.reddit.com/r/davinciresolve/comments/1wafb08/davinci_resolve_211_release_notes/",
                 )
 
--- a/client/ayon_resolve/hooks/pre_resolve_startup.py
+++ b/client/ayon_resolve/hooks/pre_resolve_startup.py
@@ -1,6 +1,9 @@
 import os
+import re
 
 from ayon_applications import PreLaunchHook, LaunchTypes
+from ayon_core import lib
+
 from ayon_resolve import RESOLVE_ADDON_ROOT
 
 
@@ -13,6 +16,26 @@ class PreLaunchResolveStartup(PreLaunchHook):
     launch_types = {LaunchTypes.local}
 
     def execute(self):
+        # Detect current Resolve version and license flavor.
+        cmd = self.launch_context.launch_args + ["-v"]
+        version_output = lib.run_subprocess(cmd)
+
+        if version_output:
+            # e.g. DaVinci Resolve Studio Version 21.1.0.0014
+            resolve_version = version_output.split("\n")[0]
+
+            is_studio = "studio" in resolve_version.lower()
+            version_match = re.search(r"(\d+)\.\d+", resolve_version)
+            major_version = int(version_match.group(1)) if version_match else 0
+
+            if not is_studio and major_version >= 21:
+                raise RuntimeError(
+                    f"AYON is not supported in {resolve_version} "
+                    "Only the studio version supports python external scripting: \n"
+                    "https://www.reddit.com/r/davinciresolve/comments/1wafb08/davinci_resolve_211_release_notes/",
+                )
+
+
         # Set the openpype prelaunch startup script path for easy access
         # in the LUA .scriptlib code
         script_path = os.path.join(RESOLVE_ADDON_ROOT, "startup.py")


client/ayon_resolve/hooks/pre_resolve_startup.py
```diff
@@ -13,6 +16,26 @@ class PreLaunchResolveStartup(PreLaunchHook):
     launch_types = {LaunchTypes.local}
 
     def execute(self):
+        # Detect current Resolve version and license flavor.
+        cmd = self.launch_context.launch_args + ["-v"]
+        version_output = lib.run_subprocess(cmd)
+
+        if version_output:
+            # e.g. DaVinci Resolve Studio Version 21.1.0.0014
+            resolve_version = version_output.split("\n")[0]
```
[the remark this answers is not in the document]
rdelillo: No it should continue if not able to detect a specific version for weird reasons.

client/ayon_resolve/hooks/pre_resolve_startup.py
```diff
@@ -13,6 +16,26 @@ class PreLaunchResolveStartup(PreLaunchHook):
...
+
+            is_studio = "studio" in resolve_version.lower()
+            version_match = re.search(r"(\d+)\.\d+", resolve_version)
+            major_version = int(version_match.group(1)) if version_match else 0
+
+            if not is_studio and major_version >= 21:
+                raise RuntimeError(
+                    f"AYON is not supported in {resolve_version} "
+                    "Only the Studio version supports Python external scripting:\n"
+                    "https://www.reddit.com/r/davinciresolve/comments/1wafb08/davinci_resolve_211_release_notes/",
```
BigRoy, line 35: I don't think we should be linking to a reddit post. Either link to a formal post by Blackmagic, otherwise remove it.

BigRoy (commented):
Seems ok, but not too keen honestly on the added startup time.

client/ayon_resolve/hooks/pre_resolve_startup.py
```diff
@@ -13,6 +16,26 @@ class PreLaunchResolveStartup(PreLaunchHook):
     launch_types = {LaunchTypes.local}
 
     def execute(self):
+        # Detect current Resolve version and license flavor.
+        cmd = self.launch_context.launch_args + ["-v"]
+        version_output = lib.run_subprocess(cmd)
```
BigRoy, line 21: Might want to see how slow this is in practice. Trying this out on command line seem to at least take a second for me. Trying this with `subprocess.run` seems to take 1.6/1.7s average. 

This would mean that any user would now 'suffer' from a similar slow-down on every launch just for those users who don't have a studio build.

How about making this a separate pre-launch hook that can be disabled in settings, enabled by default. "Validate Resolve 21+ studio version". Add a tooltip that describes that this may add a small startup time to detect whether it's indeed resolve studio, an explain why it's relevant. A studio who knows is always using Studio anyway can just disable it, to avoid the cost taken?

Or, could we even find a way that doesn't slow down the startup to begin with? 🤔 Like something that checks and notifies the user without waiting but still allowing the launch? Or?

client/ayon_resolve/hooks/pre_resolve_startup.py
tweak-wtf: i believe this is the closest we get to official
https://forum.blackmagicdesign.com/viewtopic.php?f=21&t=239823

client/ayon_resolve/hooks/pre_resolve_startup.py
rdelillo: oh god.. was pretty immediate to me. Pretty slow indeed.

I like the idea about turning that into an optional hook and caching the result. 

@tweak-wtf was suggesting we could also parse the install folder content to identify studio + version.
This would also work, on the other end it'd feel like a performance vs maintenance tradeoff to me.. (a 2 seconds reliable check once per version vs an immediate check that might break on next install if docs/readme change).

iLLiCiTiT: There will be always caviads with this. If you'd use something like REZ to launch resolve you probably have no way how to do it.

Looking for the directory content is fine if we can predict the structure and if the structure is same for each platform.

BigRoy: > If you'd use something like REZ to launch resolve you probably have no way how to do it.

It should pass on the arguments the executable, so `-v` should actually work 😄 but yes, tons of caveats to be aware of.

rdelillo: do we want to go ahead and continue this with (+ caching and mime-type) ? or hold off while we figure out something better ?
I mean I can make it so worst case would be that if the test code fails, it ignores the check and Resolve Free version opens without AYON.

iLLiCiTiT: I think we can cache it. The question is how and where? I guess using the "base" arguments hash stored to a file somewhere?

rdelillo: Implemented a cache based around the file mtime and the executable filepath.

client/ayon_resolve/hooks/pre_resolve_startup.py
```diff
@@ -12,7 +18,78 @@ class PreLaunchResolveStartup(PreLaunchHook):
...
+        is_studio = "studio" in version.lower()
+        match = re.search(r"(\d+)\.\d+", version)
+        major = int(match.group(1)) if match else 0
+        return major < 21 if not is_studio else True
+
     def execute(self):
+        # Detect current Resolve version and license flavor.
+        resolve_version = self._query_resolve_version()
+        if resolve_version and not self._is_supported_version(resolve_version):
+            raise RuntimeError(
```
iLLiCiTiT, line 87:
```suggestion
            raise ApplicationLaunchFailed(
```

iLLiCiTiT: I know we don't support integration, but shouldn't we still allow to launch it anyway? (Even if the integration won't work)

client/ayon_resolve/hooks/pre_resolve_startup.py
```diff
@@ -1,6 +1,12 @@
+from __future__ import annotations
+
 import os
+import re
+import json
 
 from ayon_applications import PreLaunchHook, LaunchTypes
```
iLLiCiTiT, line 7:
```suggestion
from ayon_applications import (
    PreLaunchHook,
    LaunchTypes,
    ApplicationLaunchFailed,
)
```

client/ayon_resolve/hooks/pre_resolve_startup.py
BigRoy: The tricky thing is that we don't really have any means to notify the user of it not supporting it - because the logs generally are not visible anywhere? 🤔 So if we weren't to raise, we don't really have a mechanism of letting the user know.

As such, I think raising is the best case, and we should make it possible to disable the hook via settings.

BigRoy: Also, I still think it makes sense to make this a separate prelaunch hook from the `pre_resolve_startup.py` that is ordered just before it.

rdelillo: Done : make its own file and optional through addon settings.

BigRoy (changes requested):
Version does not get detected correctly for me on Windows, this seems to be in cache:

```
{'mtime': 1770830676.0, 'version': '\r'}
```

server/settings.py
```diff
@@ -445,6 +445,9 @@ class ProjectDatabaseOverrideModel(BaseSettingsModel):
 
 
 class ResolveSettings(BaseSettingsModel):
+    check_compatible_ayon_version_on_start: bool = SettingsField(
```
iLLiCiTiT:
```suggestion
    check_compatible_version_on_start: bool = SettingsField(
```

server/settings.py
```diff
@@ -485,6 +488,7 @@ class ResolveSettings(BaseSettingsModel):
 
 
 DEFAULT_VALUES = {
+    "check_compatible_ayon_version_on_start": True,
```
iLLiCiTiT:
```suggestion
    "check_compatible_version_on_start": True,
```

client/ayon_resolve/hooks/pre_resolve_check_version.py
```diff
@@ -0,0 +1,99 @@
...
+    def _is_supported_version(version: str) -> bool:
+        """Free Resolve 21+ dropped external python scripting."""
+        is_studio = "studio" in version.lower()
+        match = re.search(r"(\d+)\.\d+", version)
+        major = int(match.group(1)) if match else 0
+        return major < 21 if not is_studio else True
+
+    def _should_run_resolve_check_version(self) -> bool:
+        resolve_settings = self.data["project_settings"]["resolve"]
+        return resolve_settings["check_compatible_ayon_version_on_start"]
```
iLLiCiTiT:
```suggestion
        return resolve_settings["check_compatible_version_on_start"]
```

rdelillo:
> Version does not get detected correctly for me on Windows, this seems to be in cache:
> 
> ```
> {'mtime': 1770830676.0, 'version': '\r'}
> ```

Yeah I tried to reproduce on my machine could not find the same bug as you.
I tried Resolve 20.3 20.2 21 (studio and free) on Windows, I always end-up with the correct version returned by `Resolve.exe` almost immediatly. I've also tried to test with Resolve already opened still no issue on my side.

This makes me thing the capturing `-v` output from a subprocess is not as reliable as we we hoped.

I've changed the logic to read version from shipped `ReadMe.html` instead.
Not ideal but hopefully more trustable than the executable return. I retested with all of the version above, this gives me expected results. Note that the result is not cached anymore, (reading from ReadMe or a cache takes as much time).

image

server/settings.py
```diff
@@ -445,6 +445,9 @@ class ProjectDatabaseOverrideModel(BaseSettingsModel):
 
 
 class ResolveSettings(BaseSettingsModel):
+    check_compatible_version_on_start: bool = SettingsField(
+        True, title="Check compatible Resolve version with AYON menu on start."
```
BigRoy, line 449: Possibly best to add a description too that this is about the Resolve 21+ when using the free version of resolve, etc.
```suggestion
        True, 
        title="Check Resolve license version compatibility on start",
        description=(
            "Resolve 21 dropped the external python support in the free version."
            " This validates yo are running Studio licensed build when using Resolve 21+
        )
```

Also., remove trailing dot on title since others also don't have it.

BigRoy (approved):
Works, just come cosmetics left.

